PR #141, opened 2015-01-07
Argo - hardcode harvest_type

lbesnard:
* previous version of the harvester was using the build-in context
   for the harvest_type "POSITION" ... It seems like as a build-in
   context, the harvest_type was not used and stayed empty once
   deployed. harvest_type values are now hardcoded in each iDelete
   iNewRessources component.


lbesnard:
@jonescc I modified the harvester as you suggested. Once deployed, I'll check the logs on RC. We won't know if it will have worked until the second run of the harvester though. It ran well on my machine anyway.
Cheers

jonescc:
Looks like the harvester has been using null harvest_type for platform and profile processing and 'position' for position processing.

We should cleanup these records as a part of this change using a liquibase changeset.  To do this, you can 

* delete all the existing file_harvest records which will force a reharvest of all files, or
* set the harvest_type to match the harvest type now specified in the harvester i.e.
      null harvest type profile records to 'ARGOPROF'
      null harvest type platform records to 'ARGOPLAT'
      'position' harvest type records to 'POSITION'

lbesnard:
> set the harvest_type to match the harvest type now specified in the harvester i.e. null harvest type profile records to 'ARGOPROF' null harvest type platform records to 'ARGOPLAT' 'position' harvest type records to 'POSITION'

Well it seems that is what I've done, isn't it ?

jonescc:
But this doesn't match the data so a full reharvest will be done for the new harvest types and we'll have old records hanging around for the old harvest types which will be confusing for others who may look at this in future, will take up unnecessary space and may impact performance.   

What I'm suggesting is adding a liquibase changeset that either deletes all existing file_harvest records forcing a reharvest of all files or that updates the harvest type on existing records to match the ones specified in the harvester prior to performing any processing so a reharvest is not necessary.  i.e.

    <changeSet author="xxx" id="xxx">
        <sql splitStatements="true">
            update file_harvest set harvest_type = 'ARGOPLAT' where harvest_type is null and url like '%_meta.nc';
            update file_harvest set harvest_type = 'ARGOPROF' where harvest_type is null and url like '%profiles/%.nc';
            update file_harvest set harvest_type = 'POSITION' where harvest_type = 'position';
        </sql>
    </changeSet>

jonescc:
Actually just deleting file harvest may leave data for files that have been deleted, so I think updating the harvest_type to match the new values is the best option.

lbesnard:
thanks for the explanation, I was confused. will do

lbesnard:
added this 
```sql
	update file_harvest set harvest_type = 'ARGOPLAT' where harvest_type is null and file_id = (select id from  indexed_file where url like '%_meta.nc');
	update file_harvest set harvest_type = 'ARGOPROF' where harvest_type is null and file_id = (select id from  indexed_file where url like '%profiles/%.nc');
	update file_harvest set harvest_type = 'POSITION' where harvest_type = 'position';
```

jonescc:
Need to change 'file_id = ( '  to 'file_id in ('

lbesnard:
my bad. done!

jonescc:
Need to restore argo schema from /mnt/ebs/argo-backup/argo.dump prior to running this new version of the harvester for this to be a good test of what's going to happen when we deploy to production.

jonescc:
log_dir and paramFile are added automatically for all harvesters and don't need to go in the data_bag, but yes good idea to remove visualisation and reporting entries.

PR #142, opened 2015-01-12
Modified the AATAMS_BIOLOGGING_PENGUIN harvester

xhoenner:
Schema needs to be dropped, data bags don't need to be updated. Please review ASAP so that new data can be published.


lbesnard:
@xhoenner I can have a look at it if no one else is already onto it ?

xhoenner:
@lbesnard sounds good Laurent, go for it!

lbesnard:
run well on my machine (no data to test with tho)

PR #143, opened 2015-01-12
ARGO harvest type

lbesnard:
* duplicate key fixed issue, better explained in Craig's words
    This was caused by the null value for harvest type (equality comparisons
    on null's return false) and wasn't picked up before as postgres doesn't
    check for duplicate null values in unique constraints, so every harvest,
    the harvester would check for any existing file_harvest records, never
    find any because of the null equality check and insert another record.


lbesnard:
@jonescc 
I replaced the old LB statement with the new one you suggested. Since I guess you're going to put the backup back on the harvest DB ?
I ran it on my machine, no problem

jonescc:
Still need to update 'position' records to 'POSITION' as before

lbesnard:
done. Put it in the same commit

PR #147, opened 2015-01-13
Update AATAMS_SATTAG_DM harvester to harvest files from FTP server instead of HTTP website

xhoenner:
Added tSystem component in the main job, and created a subjob to unzip all zipped files.

Schema needs to be dropped, data bags need to be updated.


mhidas:
Looks ok to me @xhoenner . Just wondering why the schema needs to be dropped. Has it changed?

PR #158, opened 2015-01-29
ABOS_SOFS_SURFACE_FLUXES - metadata

lbesnard:
* add spat update for both rt and dm
 * rename destination context


lbesnard:
change databgs

mhidas:
@lbesnard @ggalibert  This one looks like it hasn't been deployed yet.

lbesnard:
I'm onto it. will change the databags and push to RC

lbesnard:
was getting pretty confused, but the most up to date commit for this harvester is https://github.com/aodn/harvesters/commit/22903e9e2790765dbbc71413527794c81e2a5b44 which has different databags for the metadata context

mhidas:
Ah! Sorry @lbesnard. This happened because by the time I marked this PR as `to_be_deployed`, there was another one ( #174) that modified this harvester. So in fact it's that PR that needs to be deployed.

lbesnard:
it s all good.  created a PR on chef private
https://github.com/aodn/chef-private/pull/1394

once merged, I ll deploy the harvester to RC

lbesnard:
@mhidas can you please have a look at the log on nsp14 ?

mhidas:
@lbesnard Unfortunately it's failing, with errors like this:

> In finally block
> Exception in thread "main" java.lang.Error: java.lang.Error: java.lang.Error: java.lang.Error: java.lang.Error: java.lang.Error: Unresolved compilation problem: 
>         logDir cannot be resolved or is not a field
> 
>         at abos_sofs_surface_fluxes.abos_sofs_surfaceflux_harvester_1_0.ABOS_SOFS_surfaceflux_harvester.iIndexFiles_1_iFilesProcess(ABOS_SOFS_surfaceflux_harvester.java:3276)
>         at abos_sofs_surface_fluxes.abos_sofs_surfaceflux_harvester_1_0.ABOS_SOFS_surfaceflux_harvester.runJobInTOS(ABOS_SOFS_surfaceflux_harvester.java:5030)
>         at abos_sofs_surface_fluxes.abos_sofs_surfaceflux_harvester_1_0.ABOS_SOFS_surfaceflux_harvester.main(ABOS_SOFS_surfaceflux_harvester.java:4818)
> Caused by: java.lang.Error: java.lang.Error: java.lang.Error: java.lang.Error: java.lang.Error: Unresolved compilation problem: 
>         logDir cannot be resolved or is not a field
> 
> etc...

I think the problem is that the logDir context parameter hasn't been added to any of the subjobs, but they have stats & logs enabled, which are meant to write to logDir. I suggest we solve this by disabling stats & logs for the subjobs, I don't think the output from these is very useful.

mhidas:
@lbesnard See #199 for my suggested fix.

mhidas:
Also, I noticed that on 14-nsp (and on 9), the logDir is not set in the conf file. 

@anguss00 Is this one set somewhere else?

PR #159, opened 2015-02-01
ANMN_ACIDIFICATION_DM metadata

lbesnard:
* add metadata spatial extent.
 * WARNING. schema name not consistent with NRT harvester


lbesnard:
@mhidas see commit comment

mhidas:
Thanks @lbesnard . Yes, it's inconsistent, but the gain from making it consistent isn't really worth the effort right now.

PR #185, opened 2015-03-05
Fix context loading and passing configuration

mhidas:
- Add Metadata context to main job.
- Enable "Transmit whole context" for UpdateMetadata subjob.
- Disable "Implicit tContextLoad" for all subjobs (they get context values from main job).


mhidas:
Over to you @anguss00

PR #187, opened 2015-03-11
modified Destination  context

bpasquer:
@anguss00 : this should fix the  issue  with the destination context.


anguss00:
On it @bpasquer

PR #190, opened 2015-03-19
ANFOG RT update

bpasquer:
-modify context name in accordance to recomandations
-add the UpdateMetadata subjob and related Metadata context


lbesnard:
seems all good to me

PR #191, opened 2015-03-20
Standardize Destination context name

bpasquer:
the  context name is now correct@danfruehauf


bpasquer:
@danfruehauf  : I name the branch wrongly: it is meant to be ran_ctd_context. I am going to do the ran_sst change now

Issue #194, opened 2015-03-30
ANMN_NRS_DAR_YON_TS warnings in log


mhidas:
Just noticed the following warnings in the log from this harvester. I'm guessing this is due to incorrect context load setup or failure to tick "transmit whole context" somewhere?

> begin;2015-03-30 09:00:01+11:00
> Warning: Parameter "Destination_Database=>harvest?loginTimeout=1000&ssl=true&sslfactory=org.postgresql.ssl.NonValidat
> ingFactory" is a new parameter of Implicit_Context_Context
> Warning: Parameter "Destination_Port=>5432" is a new parameter of Implicit_Context_Context
> Warning: Parameter "Destination_Server=>dbprod.emii.org.au" is a new parameter of Implicit_Context_Context
> Warning: Parameter "Destination_Schema=>anmn_nrs_dar_yon" is a new parameter of Implicit_Context_Context
> Warning: Parameter "Destination_Login=>anmn_nrs_dar_yon" is a new parameter of Implicit_Context_Context
> Warning: Parameter "Destination_Password=>xx" is a new parameter of Implicit_Context_Context
> Warning: Parameter "Metadata_Username=>admin" is a new parameter of Implicit_Context_Context
> Warning: Parameter "Metadata_Password=>xx" is a new parameter of Implicit_Context_Context
> Warning: Parameter "Metadata_URL=>https://catalogue-123.aodn.org.au/geonetwork" is a new parameter of Implicit_Context_Context
> Warning: Parameter "Metadata_SpatialTable=>anmn_nrs_yon_dar_timeseries_map" is a new parameter of Implicit_Context_Context
> Warning: Parameter "Metadata_SpatialColumn=>geom" is a new parameter of Implicit_Context_Context
> Warning: Parameter "Metadata_SpatialResolution=>1" is a new parameter of Implicit_Context_Context
> Warning: Parameter "sourceDir=>/mnt/imos-t3/IMOS/opendap/ANMN/NRS/REAL_TIME" is a new parameter of Implicit_Context_Context
> Warning: Parameter "include=>.*/IMOS_ANMN_.*_NRSDAR_FV01.*|.*/IMOS_ANMN_.*_NRSYON_FV01.*" is a new parameter of Implicit_Context_Context
> Warning: Parameter "Metadata_UUID=>006bb7dc-860b-4b89-bf4c-6bd930bd35b7" is a new parameter of Implicit_Context_Context
[...]

PR #198, opened 2015-04-01
ANMN NRS DAR YON - implicit context

lbesnard:
* will hopefully fix a warning issue in log files. untick implicit context in
   metadata subjob


lbesnard:
@mhidas I think what I did should do the job. let me know what you think.

in ref of https://github.com/aodn/harvesters/issues/194

mhidas:
@lbesnard Thanks, I think that should fix it.

The one thing I'm never quite sure of is what the "Use project settings" means in practice?? (It's selected in every subjob.)
@jonescc Any idea how we can see what the project settings are?

lbesnard:
@mhidas , just a follow up. thx

mhidas:
@lbesnard Sorry, I was waiting for comments about the "Use project settings" options. Never mind.

Issue #200, opened 2015-04-13
ABOS_SOFS_SURFACE_FLUXES - deployment


lbesnard:
in ref of https://github.com/aodn/harvesters/pull/199

lbesnard:
new build is on RC. harvester is queued to run today. Please check log this eve or tomorrow @mhidas 
```
talend_log abos_sofs_fl-abos_sofs_fl
```

mhidas:
Still failing:

    Caused by: java.lang.Error: Unresolved compilation problem: 
        fileId cannot be resolved or is not a field

        at abos_sofs_surface_fluxes.harvestsofsfile_0_1.HarvestSOFSFile.tPostgresqlRow_1Process(HarvestSOFSFile.java:701)
        at abos_sofs_surface_fluxes.harvestsofsfile_0_1.HarvestSOFSFile.tPostgresqlConnection_1Process(HarvestSOFSFile.java:627)

See #203  for fix.

lbesnard:
A few things missing :
in harvestSOFSfile component in main :
 * missing intValue() for the fileId : ((Long)globalMap.get("iHarvestResources_1_FILE_ID")).intValue()
 * missing url parameter

in actual harvestSOFSfile component :
 * fileId was a string type. has to be integer
 * url context was missing
 * deploymentNo context was missing

Currently running on my machine to test that's all missing. Happy to push force the modification on this branch for you if you want me to

mhidas:
Ah, thanks @lbesnard ! All this was there originally. I don't know why it disappeared. Sure, if you've already made these changes, might as well commit them.

mhidas:
@lbesnard Ok, should be ready to redeploy now!

lbesnard:
@jonescc 

Can you please have a look at the log file on nsp14 ?
```
/mnt/ebs/talend/jobs/abos_sofs_fl-abos_sofs_fl/log/console.2015-04-20-16-20-08.log
```
basically some ' it.geosolutions.geonetwork.exception.GNServerException: Error retrieving metadata in GeoNetwork' . However if I rerun the harvester, well 
```
* iUpdateSpatialExtent - files changed: false
* iUpdateSpatialExtent - updating spatial extent: false
* iUpdateSpatialExtent - files changed: false
* iUpdateSpatialExtent - updating spatial extent: false
```
which seems to me like since it failed on the previous run, but still did something, well it doens't have to do it again.

lbesnard:
[...]

PR #203, opened 2015-04-15
ABOS_SOFS_SURFACE_FLUXES: fix missing fileId context

mhidas:
@lbesnard I hope this will fix the errors in RC. Please test.

### DIFF
--- a/workspace/ABOS_SOFS_SURFACE_FLUXES/context/Destination_0.1.item
+++ b/workspace/ABOS_SOFS_SURFACE_FLUXES/context/Destination_0.1.item
@@ -2,8 +2,8 @@
 <talendfile:ContextType xmi:version="2.0" xmlns:xmi="http://www.omg.org/XMI" xmlns:talendfile="platform:/resource/org.talend.model/model/TalendFile.xsd" confirmationNeeded="false" name="Deployment">
   <contextParameter comment="" name="Destination_Port" prompt="Destination_Port?" promptNeeded="false" type="id_String" value="5432"/>
   <contextParameter comment="" name="Destination_Login" prompt="Destination_Login?" promptNeeded="false" type="id_String" value="abos_sofs_fl"/>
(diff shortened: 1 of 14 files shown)


lbesnard:
@mhidas funny, I did the modification, but it almost look that its you who did it. wanna have a look and merge ?

lbesnard:
@mhidas can you please review and merge so I can build it on jenkins and run it asap on nsp14 ? cheers

mhidas:
@lbesnard We're getting there... you fixed the issues you pointed out in #200, but then shouldn't `fileId` have the same type in the `DeleteSOFSFileData` subjob?

Also, `fileId` should actually be of Long type, according to the type of the `file_id` column that gets its value from `fileId`. In that case, do you still need intValue() ?

lbesnard:
> Also, fileId should actually be of Long type, according to the type of the file_id column that gets its value from fileId. In that case, do you still need intValue() ?

it makes sense (if I understood properly). Didn't get any error tho, so maybe talend does some type conversion on its own

> shouldn't fileId have the same type in the DeleteSOFSFileData subjob?

will have a look

lbesnard:
@mhidas  done your suggestions (both)

lbesnard:
@mhidas  FYI :
```In Java, the language itself determines how the certain types are represented. The order is: byte 8 bits, short 16 bits, int 32 bits, long 64 bits. In Java, all types are signed, there are no unsigned versions.Feb 21, 2013```

And
http://stackoverflow.com/a/15005226/4294174


In other words, 2^32 = 4294967296 for unsigned values. I think we're safe to use INT as a type in java before going over the limit of files to harvest.

mhidas:
@lbesnard My point was mainly about being consistent with the database. However I've just realised that even though the `file_id` columns in the `measurement` and `file_metadata` tables in the db are of type "bigint", the `id` they point to in `indexed_file` is only "int" anyway. So all I can say now is "Whatever...."

lbesnard:
@mhidas  can this be reviewed/merged please so I can deploy it ? cheers

lbesnard:
thx. will deploy

PR #205, opened 2015-04-21
Abort by default if no files present in source directory to prevent all ...

jonescc:
...data being removed from the database

This pull request will change the behaviour of iIndexFiles in all harvesters to abort if there are no files in the source directory.  This is to prevent all data from being removed by harvesters e.g.  by iDeleteResources components due to system errors such as the sshfs filesystem problem that occurred recently for SRS data.

It sets the 'Generate Error if no file found' option on iIndexFiles components in these harvesters.

If you are working on any of these harvesters, please ensure you turn this option on in your version to ensure these changes are not reverted when your changes are merged.

If you think any of these harvesters need to handle empty directories - let me know and I'll remove them from the changes below


anguss00:
Awesome @jonescc :+1:

lbesnard:
@aodn/emii-ops  if no-one is doing any changes on harvester, and is currently working on a branch, I'd like to merge this asap, otherwise you might end up having some conflicts, and I suspect that doing a rebase against master with talend could be tricky

mhidas:
@lbesnard No unmerged branches here. As far as I'm concerned, you can go ahead.

PR #207, opened 2015-04-28
ARGO - fix erros

lbesnard:
* 20000+ files were having issues.
 * there was a probleme with the CALIBRATION_DATE and HISTORY_DATE variables which
   were empty, but returning a wrong value of string, instead of returning null. This
   is wrong, and probably due to the NetCDF java lib (ncdump worked though). This was
   handled by putting TRY CATCH in the parseArgoDate talend utils.
 * The type of two different columns (in tmap7 and tmap9 for ref) were of the type
   varchar(16) which was becoming too small for some new files. Handled it by changing
   in liquibase the type to text. Views had to be droped with CASCADE. The creation of
   views was changed in liquibase to add a runAlways = true which will always drop a
   view if exists and recreates it. This ways, the liquibases changeset are way more
   clean
 * remove the liquibase changeset to get rid of functions, since this is now unacessary
 * put all liquibase components into a subjob, put at the start of the harvester.


lbesnard:
worked and tested on my pobox. ready to be reviewed and merged

lbesnard:
in ref of this : https://github.com/aodn/harvesters/issues/213#issuecomment-98862116

which needs to be fixed before merging this PR. All data will have to be reharvested from an empty schema

lbesnard:
**update 20/05/2015** : the schema doesn't have to be dropped for this PR. Merging and testing on RC should be enough. however, once some work will have been done regarding string values of QAQC (mentionnend above 

> in ref of this : #213 (comment)
)

the schema will have to be dropped because most of the data seems to have wrong values of QAQC

lbesnard:
@ggalibert  just wanted to try to make sure. I used a sample of data (argo_testing.tar.gz on my home on nsp10). It should fail with the current prod version. But with the new one it should succeed. I ran it on a new schema and was also fine
```sql
select drop_objects_in_schema('argo')
```

lbesnard:
@jonescc 
```
/mnt/opendap/1/IMOS/opendap/Argo/dac/csiro/5900034
```

Issue #182, opened 2015-03-02
ANMN_NRS_CTD_PROFILES harvester failing on new files


mhidas:
Failing to harvest 4 recently published files. Messages in console.log look like this:

    In finally block
    Exception while processing /mnt/imos-t3/IMOS/opendap/ANMN/NRS/NRSROT/Biogeochem_profiles/IMOS_ANMN-NRS_CDEKOSTUZ_20100520T020500Z_NRSROT_FV01_3015.0-1-SBE19plus-45_C-20140130T091925Z.nc
    java.lang.NullPointerException
        at anmn_nrs_ctd_profiles.dataharvest_0_1.DataHarvest.iDeletedResources_1Process(DataHarvest.java:1477)
        at anmn_nrs_ctd_profiles.dataharvest_0_1.DataHarvest.iPostgresqlDbUpdate_1Process(DataHarvest.java:1326)
        at anmn_nrs_ctd_profiles.dataharvest_0_1.DataHarvest.runJobInTOS(DataHarvest.java:3698)
        at anmn_nrs_ctd_profiles.dataharvest_0_1.DataHarvest.runJob(DataHarvest.java:3526)
        at anmn_nrs_ctd_profiles.anmn_nrs_ctd_profiles_harvester_0_1.ANMN_NRS_CTD_PROFILES_harvester.tRunJob_2Process(ANMN_NRS_CTD_PROFILES_harvester.java:1319)
        at anmn_nrs_ctd_profiles.anmn_nrs_ctd_profiles_harvester_0_1.ANMN_NRS_CTD_PROFILES_harvester.tRunJob_1Process(ANMN_NRS_CTD_PROFILES_harvester.java:1202)
        at anmn_nrs_ctd_profiles.anmn_nrs_ctd_profiles_harvester_0_1.ANMN_NRS_CTD_PROFILES_harvester.runJobInTOS(ANMN_NRS_CTD_PROFILES_harvester.java:3281)
        at anmn_nrs_ctd_profiles.anmn_nrs_ctd_profiles_harvester_0_1.ANMN_NRS_CTD_PROFILES_harvester.main(ANMN_NRS_CTD_PROFILES_harvester.java:3085)

The files exist, have the same permissions as all the other files in the directory, and I can't see anything obviously wrong with them. I've seen this `NullPointerException` error before, but I can't remember what the problem was.

mhidas:
This is a duplicate of https://github.com/aodn/content/issues/80

PR #212, opened 2015-04-29
ANMN_NRS_CTD_PROFILES: fix db connection for delete

mhidas:
This harvester has been failing (see #182 and https://github.com/aodn/content/issues/80) because it's been trying to delete data for files that have been removed from the filesystem, but the db connection parameters were incorrectly set for the component that does this. It was trying to "use existing connection", but there wasn't one in that subjob.

This should fix it. I don't have the PO VM set up to test this, so whoever reviews this, please do a test run, delete some files, then test run it again to make sure it works. Thanks!


mhidas:
@bpasquer @lbesnard @xhoenner Please review & merge this. It's a small change so shouldn't take long. Dan reckons we can just test it in RC so you don't even need to run it.

PR #215, opened 2015-05-04
Update wdb observed data

bpasquer:
Major update. 
The harvester harvest new set of data. Data Bag need to be updated.


lbesnard:
@bpasquer 
suggestions :

 * only have sourceDir in source context
 * then every single file context would contain only the filename and could be built in contexts. In your different tfileInputDelimiter, just create a string such as 'context.sourceDir + "/" + context.SourceMBT_data. I think this should work, maybe worth asking @jonescc before though

This makes the deployment so much easier for you and for me. The databags don't have to be changed anymore, you just have to reference all the files from sourceDir

lbesnard:
@bpasquer 
I could not run the harvester properly on my machine, I changed the sourceDir to 
```
/mnt/imos-t4/IMOS/archive/eMII/TALEND_harvester/WODB/
```
But I think some files are missing such as /mnt/imos-t4/IMOS/archive/eMII/TALEND_harvester/WODB/ObsData_textfiles/CTD_obs.txt 

Also, one remark, usually, the sourceDir doesn't have a slash at the end. So in the tFileInputDelimited, it's usually better to put 'context.SourceDir+ "/ObsData_textfiles/XBT_obs_all.txt"' (sorry). Saying this because it's likely someone (even me if I forget) would do the mistake while writing the databags

lbesnard:
I have many errors such  as 
	Type mismatch: cannot convert from Float to int

lbesnard:
The problem is in uor and mbt subjobs. In the tmap, for example, cast_id goes from float to int

lbesnard:
@bpasquer I'm updating your harvester to add a slash before all 'ObsData_textfiles/*.txt' in tfileinputDelimiter. The reason of this is because the sourceDir in the chef context doesn't have a slash at the end :

```
cat /mnt/ebs/talend/jobs/wodb-wodb/etc/wodb-wodb.conf | grep SourceDir
SourceDir=>/mnt/imos-t4/IMOS/archive/eMII/TALEND_harvester/WODB
```

Also, in your next PR once you'll clean everything, can you also change SourceDir to sourceDir to match the other harvesters ?

I have one issue with your XBT subjob, it can't find the following file :
/mnt/imos-t4/IMOS/archive/eMII/TALEND_harvester/WODB/ObsData_textfiles/XBT_obs_data.txt

do XBT_obs_all_data.txt and XBT_obs_all.txt have to be respectively renamed to XBT_obs_data.txt and XBT_obs.txt on the filesystem ? tell me and I'll do it

bpasquer:
Ok, I have changed the XBT_obst file name.

lbesnard:
oh you didn't have to do it, I would have done it as I'm already making changes. So no need to change files on the filesystem then ?

bpasquer:
No, file names on the filesystem has to remain unchanged.

lbesnard:
I rebased all the 7 last commits into 1. merging now, worked fine on my machine

PR #216, opened 2015-05-05
ANMN acidification mooring - add dissolved oxygen variable

lbesnard:
* the DOX variable was added to all delayed mode netcdf files back in
   August 2014. This is now added to the harvester and db views
 * the metadata record needs to have the DOX1 variable added
    mole_concentration_of_dissolved_molecular_oxygen_in_sea_water.
    units: micromol/l
 * the schema needs to be dropped since the files have been the same for
   almost a year, so they need to be re-harvested
 * the geoserver filter will have to be updated afterwards in order to filter
   on the dox1 variable


lbesnard:
> the metadata record needs to have the DOX1 variable added mole_concentration_of_dissolved_molecular_oxygen_in_sea_water. units: micromol/l

this will be for your @atkinsn 

> the schema needs to be dropped since the files have been the same for almost a year, so they need to be re-harvested

would you be able to help me on this @anguss00 

This PR is an example of why it would be valuable to have geoserver, talend and geonetwork in one PR @danfruehauf @jkburges

atkinsn:
Done @lbesnard

bpasquer:
You're changes look all good to me but, i spotted a few details that could be improved in the harvester. For example the quality_control parameter are declared as float instead of being character(1). I'll merge your PR and do these changes cause you know what: i created this harvester a year ago....

PR #218, opened 2015-05-05
Hawksbill harvester: add GPS data

xhoenner:
Shouldn't need to drop the schema, data bags need to be updated (new metadata context group, get UUID from harvester).


lbesnard:
@xhoenner when trying to run it on my machine, i get a missing jar warning for 3 jars. Do you have something special set up >?

xhoenner:
@lbesnard these are the drivers to read the Microsoft Access files, not sure where they are stored on 9-nsp, any idea @danfruehauf?

lbesnard:
ok found them in talend/jobs/aatams_sattag_dm-aatams_sattag_dm/java/lib/

lbesnard:
the question is where do you put them for a non built version ?

lbesnard:
Tip how to add *.jar . see point 2) 
http://www.vikramtakkar.com/2013/03/importing-external-jar-or-library-files.html

works well on my machine.only have some warnings such as 
```
WARN  Unsupported data type BINARY for index, making read-only
```

So just to make sure, you have 1 wms, and 6 wfs ? so 6 different metadata records ?

lbesnard:
Just one comment, I have the feeling it should be changed. In the main job, the metadata_..._GPS contexts are __built-in__ and not taken from the context.

lbesnard:
updated databags here https://github.com/aodn/chef-private/pull/1437

xhoenner:
> Tip how to add *.jar . see point 2) 
http://www.vikramtakkar.com/2013/03/importing-external-jar-or-library-files.html

Thanks Laurent, I'll try that next time.

> So just to make sure, you have 1 wms, and 6 wfs ? so 6 different metadata records ?

Correct, I use the _map table to create 5 additional WMS in Geoserver though.

lbesnard:
> Just one comment, I have the feeling it should be changed. In the main job, the metadata_...GPS contexts are __built-in_ and not taken from the context.

all good now.

PR #219, opened 2015-05-06
Changed measurements.idx index to include timezone

xhoenner:
No need to update the data bags, nor to drop the schema. 

This will fix https://github.com/aodn/content/issues/93


lbesnard:
looks good. will see how nagios rc will behave. @danfruehauf to let me know if the change worked. I'll run it on rc tonight

PR #220, opened 2015-05-08
Changed indexes to improve Geoserver performances

xhoenner:
@danfruehauf that should fix the existing issues on nagios.


danfruehauf:
That also gets rid of a few CSV files. Which is probably good, but better to usually have those in a different commit.

PR #221, opened 2015-05-08
ANMN WAVE - new harvester

lbesnard:
* new harvester to process ADCP data (adcp + awac which is an adcp)
 * no metadata record yet


lbesnard:
@atkinsn I don't think you can review this harvester, but it's just to let you know we need to create a metadata record with Seb

bpasquer:
i just have a few comments @lbesnard  but overall it 's good.
- All the measured parameters are declared as real in the adcp and awac measurements table but when looking at the tPostgresql Output componement of the harvest_adcp and  harvset_awac  jobs these same parameters are declared either as FLOAT4 (real) or FLOAT8(double). We usually store all parameters in the database as real .
-in the delete_awac component : tick 'Transmit the wholecontext'
-i am not sure the sourceDir path is correct as i tried it and failed to find any netcdf file . Is  the path missing slashes to indicate in which subdirectory to look , like  this : '.*/Wave/.*FV01.*.nc' ?

bpasquer:
Another comment regarding the way you deal with ADCP and AWAC data  :  don't you need to apply a custom query in the iDeleteResource to delete separately awac and adcp data the same way you do in the iHarvestResources?

lbesnard:
> Another comment regarding the way you deal with ADCP and AWAC data : don't you need to apply a custom query in the iDeleteResource to delete separately awac and adcp data the same way you do in the iHarvestResources?

This is actually handled :
* in delete awac
```sql
"delete from timeseries where file_id = " + context.fileId + "; 
delete from visualisation_awac where file_id = " + context.fileId + "; 
 delete from measurement_awac where file_id = " + context.fileId + ";"
```
* in delete adcp

```sql
"delete from timeseries where file_id = " + context.fileId + "; 
delete from visualisation_adcp where file_id = " + context.fileId + "; 
 delete from measurement_adcp where file_id = " + context.fileId + ";"
```

is this what you meant ?

lbesnard:
@bpasquer done the changes and tested it. all good for me. Rebased it. Ta

PR #226, opened 2015-05-13
SRS OC slow layers

lbesnard:
* change all views to subsitute regexp with like.
 * add index on timeseries


danfruehauf:
That's looking good. Thanks.

PR #227, opened 2015-05-13
Aims ctd idx

bpasquer:
@lbesnard  i added indexes in liquidbase, modified the data view to output url and size and changed iPostgresqloutput component
Runs fine on my machine.
please review


lbesnard:
@bpasquer , see my commit message. deploying now to RC

PR #232, opened 2015-05-14
FAIMMS - fix harvester - exception with file deletion

lbesnard:
* error with processing of deleted files. Likely to be caused by and harvest type
   not being different between QAQC and NO QAQC data. Harvest type is different now
   in iharvest and idelete
 * Add SQL update query in liquibase to update harvest_type to 'qaqc' and 'noqaqc'
   depending of the file URL


danfruehauf:
Hope it runs without a need for a full re-harvest.

PR #242, opened 2015-05-27
AODN CSIRO - METADATA UPDATE

lbesnard:
* modify uuid value for CTD and ADCP layers
 * tick modify always the catalog value at each run. As the UUID has changed,
   this is the only way to update the new record
 * create another context group for metadata URL Password and Username instead
   of leaving it in the ADCP context. Did the mod needed for this


lbesnard:
@atkinsn Did the change.

PR #243, opened 2015-06-09
ANMN_NRS_BGC: column renamed in pytoplankton table

mhidas:
CSIRO have renamed the column BIOVOLUME_ML_PER_L to BIOVOLUME_UM3_PER_L in the phytoplankton table. The change in units means the data change as well, so we clear all the data out before renaming the column (the harvest will fill in the new values).


mhidas:
@lbesnard @bpasquer Could one of you please review and merge this fairly soon? CSIRO will make this change tomorrow (Wednesday). It would be good to be able to test this version in RC straight away and hopefully deploy to prod before the previous version starts failing.

workspace/ANMN_NRS_BGC/code/liquibase/anmn_nrs_bgc.changelog.xml
```diff
@@ -488,7 +498,7 @@ CREATE VIEW anmn_nrs_bgc_plankton_phytoplankton_data as
     p."CAAB_CODE",
     p."TAXON_START_DATE",
     p."CELL_PER_LITRE",
-    p."BIOVOLUME_ML_PER_L",
+    p."BIOVOLUME_UM3_PER_L",
```
lbesnard: this harvester will fail. The changelog id is the same, with different sql query in. If you want to keep the same id, why not doing (if not used in a view)
```
<changeSet author="talend" id="phyto_taxon_changelog-1" runAlways = true >
       <sql>	...	       <sql>
```

lbesnard:
> CSIRO have renamed the column BIOVOLUME_ML_PER_L to BIOVOLUME_UM3_PER_L in the phytoplankton table. The change in units means the data change as well, so we clear all the data out before renaming the column

does that mean the metadata header needs to be modified as well ?

workspace/ANMN_NRS_BGC/code/liquibase/anmn_nrs_bgc.changelog.xml
mhidas: You're looking at the wrong changeSet. The one I've changed starts with

    <changeSet author="talend" id="plankton_phytoplankton_data_view" runOnChange="true">

mhidas:
@lbesnard I don't think this one has a metadata header yet.

workspace/ANMN_NRS_BGC/code/liquibase/anmn_nrs_bgc.changelog.xml
lbesnard: I dont understand. I'm just looking at the diff of this xml, and I definitely see a diff on 
```
<changeSet author="talend" id="phyto_taxon_changelog-1">
``` 
and 
```
<changeSet author="talend" id="phyto_biovolume_column_change">
```

mhidas: It just looks that way because GitHub cuts out unchanged lines in between. Look at the line numbers on the far left. You'll also find an "Expand" button there to view the hidden lines.

lbesnard: my bad

PR #246, opened 2015-06-10
ANMN_NRS_BGC: fix uuid in spatial update

mhidas:
After running this harvester in RC with the spatial update component enabled for the first time, I noticed that the UUID was set incorrectly for the zooplankton abundance metadata record. Easy to miss, it only differs from the biovolume one by one character!

This same change will need to be applied in the data  bags for it to take effect!


lbesnard:
is the previous uuid part of the same harvester ? if not, the harvester of where this uuid belongs to would have to be modified if the iupdate component does't  'update always' the record

mhidas:
@lbesnard Yes, the incorrect uuid is one of the other products handled by this harvester, so just updating this one will fix everything.

lbesnard:
good, I'll merge this PR tomorrow (which is actually not really important), and change the uuid from the databgs at the same time

PR #247, opened 2015-06-12
Support new skos/rdf format

jonescc:


danfruehauf:
@jonescc Can I please get a bit of a background and description for that?

PR #248, opened 2015-06-23
ANMN_WAVE: set correct Metadata_UUID

mhidas:
This harvester is running in production, but the spatial extent in the metadata record hasn't been updated. Turns out there's a typo in the metadta UUID value in the data bags. I've set the correct value here.

@lbesnard Please merge and update the data bags in RC & prod. Thanks.


lbesnard:
done

PR #250, opened 2015-07-06
Argo - solving issues with float 2902088

lbesnard:
* a parameter names 'PARAMETER' had a too little size defined in its table.
   modified this from type charvar(16) to text <-- liquibase modification


lbesnard:
solve 277 files conflicts. I'll probably merge it myself (in less that 10')

PR #251, opened 2015-07-07
Fix up aatams acoustic

xhoenner:
Data bags for Origin context group have changed to:
server: aatams.emii.org.au
port: 5432

Please @lbesnard review ASAP so that it can be tested on rc today.


lbesnard:
not necessary a bad idea to rebase many commits into one when all linked to a same issue

otherwise the snail freaked me out !

PR #257, opened 2015-07-15
AUV VIEWER TRACK - modify input files

lbesnard:
* the process of one track was divided between two files DATA*.csv and METADATA*.csv
   this created issues if one file out of the two was 'touched' or modified requiring
   a complexe workflow in talend which actually didn't work
 * Both files are now appended into one (modified with quick bash script, and added to
   the data-services repo) called DATA_[DIVE_NAME].csv
 * The harvester had to be change to take this into consideration, and is now more robust
 * the databags can also be simplified, as some context disapeared
 * probably few other things


lbesnard:
see  linked PR https://github.com/aodn/data-services/pull/181

PR #261, opened 2015-08-06
improve jenkins build trigger

danfruehauf:
* allow rebuild (pass --rebuild)
 * ignore some blank harvesters (and report them)

### DIFF
--- a/bin/jenkins-trigger-helper.sh
+++ b/bin/jenkins-trigger-helper.sh
@@ -6,26 +6,57 @@ get_harvester_job_name() {
 }
 
 rebuild_artifact() {
-    local artifact_directory="$1"; shift
+    local artifacts_directory="$1"; shift
     local project_directory="$1"; shift
     local harvester_name="$1"; shift
     [ x"$harvester_name" = x ] && return 1
 
     local artifact_path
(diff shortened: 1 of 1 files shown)


bin/jenkins-trigger-helper.sh
```diff
@@ -22,6 +22,12 @@ main() {
     local artifact_directory="$1"; shift
     local projects_directory="$1"; shift
     local output_directory="$1"; shift
+    local -i rebuild_all=0
+
+    if [ x"$1" != x ] && [ "$1" == "--rebuild" ]; then
```
jonescc: Should we name $1 - I'm guessing its a harvester name?

jonescc: OK - I can see now its the rebuild option

danfruehauf: You *can* convince me to factor that into using `getopts`. It'll be many more lines though.

jonescc: I was just about to ask about getopts!  At least explaining what the options are to this script at the top of it would be beneficial - getopts would make it more standard/reviewable?

danfruehauf: Yeah, let me do it then, shouldn't take long! :)

danfruehauf:
@jonescc now with `getopts`.

PR #272, opened 2015-08-18
ANMN_NRS_BGC and SOOP_AUSCPR: increase WFS timeout limit

### LINKED ISSUE #271: ANMN_NRS_BGC and SOOP_AUSCPR intermittent timeout issues
ANMN_NRS_BGC and SOOP_AUSCPR both fall over occasionally with timeout errors trying to connect to the CMAR server.  The issue is with the CMAR server occasionally taking too long to respond to WFS requests.

mhidas:
This is needed to cope with the sometimes slow response from the CMAR Geoserver (well, slower than 3 seconds).

Should fix #271.


danfruehauf:
Can those values be externalized? Controlled as a context parameter?

mhidas:
@danfruehauf Good idea -- why didn't I think of that? Ok, I'll do this and push changes here.

mhidas:
Ok, timeout limit has been externalised.

When redeploying, the new context variable `WfsTimeout` will need to be added to the databags for both of these harvesters.

mhidas:
@danfruehauf @ggalibert @bpasquer @xhoenner Could someone please review and merge?

These harvesters have been failing for about a month. This should get them running again.

PR #274, opened 2015-08-20
AODN_RAN_CTD :fix index

bpasquer:
- in liquibase : drop index before creating.
 - change database name in context Destination (for local application)


bpasquer:
Quick fix please review  and redeploy on 14 and prod. No change in databags

xhoenner:
Looks all good to me @bpasquer, I've tested this fix using a backup of the aodn_ran_ctd schema on 14-nsp to make sure there isn't any Liquibase conflict anymore. @ggalibert deploy it bro!

PR #279, opened 2015-08-28
MHL_SST : corrected thredds url for file download in map view

bpasquer:
super quick change .please review  and redeploy . no change in databags


xhoenner:
This harvester also needs to be re-deployed @dnahodil?

PR #280, opened 2015-08-28
RAN CTD : corrected thredds url for file download in map view

bpasquer:
exact same as mhl sst. Pls review and deploy.


xhoenner:
Could you re-deploy this harvester @dnahodil please?

PR #285, opened 2015-09-09
SOOP SST - remove autocommit

lbesnard:
* use to commit a transaction using both tpostgrescommit and tcommit at
   the same time leading to some errors. removed autocommit


lbesnard:
ref: 
https://github.com/aodn/content/issues/125

PR #286, opened 2015-09-09
ARGO - add map layer

lbesnard:
* add map layer as a hard view, instead of using the sql capability of
   geoserver


lbesnard:
Bundled with PR https://github.com/aodn/geoserver-config/pull/278

Should fix issue https://github.com/aodn/content/issues/130

PR #287, opened 2015-09-10
 ran_ctd:add filter on position_qc in  map view

bpasquer:
-change query to create map
plus local change of destination database name


bpasquer:
the major change is a one line fix . Pls review and merge

PR #288, opened 2015-09-10
AUV VIEWER - geometry modification

lbesnard:
* the AUV tracks were displayed weirdley by geoserver in the last few
   months. Suggestions about modifying the buffer, and then problem when
   the start and end point were not in the same tile. An indea was put
   to change from linestring to multilinestring.
 * Replaced the auv_tracks_vw view geometry with ST_dump(geom) which
   appears to fix it on testing


lbesnard:
@dnahodil  wanna merge ?

dnahodil:
Hell yeah, I do

PR #289, opened 2015-09-14
AUSTRALIAN PHYTOPLANKTON DATABASE - NEW harvester

lbesnard:
* New harvester to get Phyto data from CMAR geoserver
 * sWFsInput component can't deal with NULL values from a geoserver
   link. Instead of using this component, the data is downloaded in the
   TALEND cache and directly loaded as a CSV.
 * URL's provided to download data from geoserver was shortened with
   google shortener, since the URL contains special character the TALEND
   component could'nt deal with
 * create a view to join both tables on project_id


lbesnard:
harvester requested by @ktattersall .

xhoenner:
Okay Loz, the harvester ran successfully on my machine. Some comments below before merging this PR:

* When running the harvester, I get the following warnings, is that expected behaviour?
screen shot 2015-09-14 at 3 59 56 pm

* For some reason the adp_data table ends up with two GIST indexes on the same column.

* Might be worth putting an index on the "PROJECT_ID" column of the adp_data table so that a SELECT query on the aus_phyto_db_data view returns data faster.

* In the data_harvest subjob, I'd suggest pulling out the iUpdateSpatialExtent component and create a separate subjob, as recommended here: https://docs.google.com/document/d/1-E05wDMvI_Yy2Zz9m3D8DCWUjjkhPDJFm_jr9TvbUgc/edit#heading=h.onsdd540oals

* In the same subjob, there's a context item called 'WfsServerURL' that's not part of a context group, to delete?

ktattersall:
Thanks @xhoenner

lbesnard:
@xhoenner did everything except for the first point as I have no idea where this is coming from

PR #291, opened 2015-09-17
fix to issue with minimap no displaying right spatial extent:change o…

bpasquer:
…ption for update of Saptial extent to always in iUpdateSpatialExtent component


bpasquer:
@lbesnard  : could you please take care of this one and deploy it pls.

PR #295, opened 2015-09-21
Created new harvester for snow petrel data

xhoenner:
Data bags to be created (UUID to be provided by Nat), new schema and role needed. 

SourceDir: 10-nsp-mel.emii.org.au:/mnt/imos-t4/IMOS/archive/eMII/TALEND_harvester/AATAMS/snowpetrel


lbesnard:
@xhoenner I can't find aatams_gls_metadata.csv to put in the context

lbesnard:
* context:
DestinationDb to Destination

* index 
maybe add an index on animal id since you're ordering on animal id and time ? Not sure if it would do anything as animalId is already a pkey

* view _data to _map ? can't remember what s the convention. For my last harvester I rechanged the view to _map ...

works fine otherwise, GJ

xhoenner:
Thanks @lbesnard for your comments. I've addressed the first two in https://github.com/aodn/harvesters/commit/b7d367b894c8d4325ffb98ac54ba8c16e0d7b0db. The view _data (3rd comment) is correct, there's a _map table to be used for the WMS layer.

bpasquer:
Just two comments because i like being picky 
-in the  aatams_biologging_snowpetrel_map why the start_date and end_date timestanmp without timezone?
job Harvest_snow_petrel_metadata the Deployment_longitude should be declared as double not float.

xhoenner:
@bpasquer Changed start_date and end_date as requested. Haven't changed Deployment longitude as the type for this particular field is handled in Liquibase anyway.

lbesnard:
good to be merged ??

bpasquer:
yes you can

PR #305, opened 2015-09-30
Added component to delete erroneous files in AATAMS_SATTAG_DM harvester

xhoenner:
Changes to fix https://github.com/aodn/content/issues/141, https://github.com/aodn/content/issues/139, and https://github.com/aodn/content/issues/134

Data bags need to be updated: new context 'file_remove' pointing to bash script 'to_delete.sh' in aatams_sattag_dm directory on 9 and 14-nsp 

Prior to deployment, @danfruehauf could you please copy that .sh file from my home directory into /mnt/ebs/data/IMOS/archive/eMII/TALEND_harvester/AATAMS/aatams_sattag_dm on 14-nsp?


danfruehauf:
@xhoenner Done.

xhoenner:
Great, thanks @danfruehauf.

danfruehauf:
It's pretty ugly here @xhoenner, that the harvester is calling a bash script. I might even say this is dangerous. I think we should be seeking a different solution here.

xhoenner:
How would you do it then? I need to be able to tell Talend not to process certain files from outside the harvester and update that list of files to exclude on occasions. Unless you know of a way to check the validity of a MS Access file?

The bash script only looks at the specified file paths so I don't understand why this is dangerous.

danfruehauf:
Having a talend harvester trigger an executable is not ideal and can be considered dangerous. This is not good practice.

You can use the pipeline processing to move files to this location only if they are valid MS access files, otherwise - push them to an error directory for further inspection. Where do those files come from anyway?

xhoenner:
The pipeline processing hasn't been implemented yet for AATAMS SATTAG MS Access files. 

Those files are copied across by the harvester from the tag manufacturer's FTP server using lftp (configured in Talend). Moving the files to an error directory won't solve the current issue as the next time the harvester runs it will run the lftp command which will copy again the files that have been moved.

danfruehauf:
I understand. In that case I don't think the harvester should be running lftp. Probably lftp should be run externally (using cron or such) and then the files moved to where they need to be.

PR #308, opened 2015-10-08
Modify acorn_radial_nonqc to use manifest harvesting components

danfruehauf:


danfruehauf:
Made obsolete by #343.

PR #309, opened 2015-10-09
AUS PHYTO DB - add missing columns to view

lbesnard:


lbesnard:
an easy PR (tested) to runAlways , and to add many missing columns 

if a nice soul wants to increase his/her carma before the we

PR #315, opened 2015-10-27
Event driven ABOS_SOTS

danfruehauf:


mhidas:
Seems to work ok.

danfruehauf:
Lets merge all those branches then and push it to production.

PR #320, opened 2015-10-29
Event driven SOOP_BA

danfruehauf:


danfruehauf:
Still untested.

danfruehauf:
Now tested. Works well!

PR #323, opened 2015-11-03
Don't add base path to assignment expressions which typically don't n…

jonescc:
…eed it


danfruehauf:
Nice. I hope that'll cover most cases (if not all).

lbesnard:
@jonescc what shall be done with this ?

jonescc:
Well the migration is over so we don't need this anymore

PR #326, opened 2015-11-26
SOOP TRV - pipeline

lbesnard:
* files are modified so they pass CF and IMOS checker, ned to take this
   into account
 * modified harvester so it uses the pipeline process
 * bug fixing, removed some useless indexed
 * Need to remove objects from schema


lbesnard:
needs to be merged first:
https://github.com/aodn/data-services/pull/255
https://github.com/aodn/geoserver-config/pull/302

PR #330, opened 2015-12-08
SRS SST - fix time harvest

lbesnard:
* time was previously harvester from time_coverage_start, which is
   different from the time variable
 * All files are reprocessed, by puting a liquibase statement which
   updates the last modified date value to epoch (1970,1,1)
 * iNewRessources replaced wit iModifiedRessources


lbesnard:
this should fix https://github.com/aodn/content/issues/149

xhoenner:
Just for the record: it looks all good, except that iModifiedRessources must be replaced by iHarvestResources.

lbesnard:
Thanks @xhoenner  did the modifications

PR #341, opened 2015-12-21
Fix up _summary_data materialized views as they are not supported by Geoserver

xhoenner:
This PR addresses https://github.com/aodn/content/issues/159.

@danfruehauf if you're happy with this, could you please merge and deploy to RC?


danfruehauf:
So... Is it just a view now?

xhoenner:
Yes it is, a view based on the materialised view. I tested that new config on my VM, both for Geoserver and for db performance purposes.

PR #344, opened 2016-01-06
SOOP TRV - fix db blow up

lbesnard:
* Missing GROUP BY in sql query which caused data to be quadriplicate
   (?) and really slow to process
 * fix issue in display of track colour for different vessels
 * truncate measurements_merged_data table with liquibase


lbesnard:
TO DO:
- VACUUM the schema after the harvester has finished running could be great
- re-pushing all the files already on S3 to INCOMING_DIR
- pushing the ones already in the ERROR DIR.

Alternatively, the harvester could run on a new schema

danfruehauf:
> TO DO:

TODO is probably not the best name to describe that. Perhaps go with "outage plan" or something like that...

danfruehauf:
Suggested fix for https://github.com/aodn/content/issues/165

PR #345, opened 2016-01-06
GSLA trigger

danfruehauf:
GSLA event driven harvesters


danfruehauf:
So... not quite ready for merging. I want to modify the harvester to run on `.nc` files instead of `.nc.gz`. The incoming handler will feed the harvester the `.nc` file, but it shall be indexed as a `.nc.gz` file. Then uploaded to S3 as a `.nc.gz`.

danfruehauf:
@ggalibert Second thought, we can merge this and deploy to the test environment - verify everything works as expected. Prod can continue working as usual.

PR #348, opened 2016-01-07
ABOS_CURRENTS: handle missing timestamp attributes

mhidas:
Some files are missing global attributes like `time_deployment_start` etc... and have been crashing the harvester. This will just quietly set them to null.


ggalibert:
Deployed to RC.

mhidas:
Thanks @ggalibert!

mhidas:
@ggalibert Could you please deploy it to prod too? I've tested it in RC and it's good to go.

danfruehauf:
@mhidas Will be deployed soon.

PR #353, opened 2016-01-11
Event driven AATAMS_SATTAG_DM

danfruehauf:
Reverts aodn/harvesters#351


danfruehauf:
OK, this PR works perfectly. At least so I believe.

@xhoenner We should experiment with that tomorrow + deploy.

xhoenner:
Simplified harvester by removing 2nd index and harvest components. I tested it out @danfruehauf so it's ready to be deployed.

danfruehauf:
> Simplified harvester by removing 2nd index and harvest components. I tested it out @danfruehauf so it's ready to be deployed.

Thanks. That's good. I guess we'll deploy it first thing tomorrow morning then.

Are you generally happy with the rest of the workflow?

xhoenner:
Sounds good to me. Yes the new workflow looks good @danfruehauf, nice work putting all that together!

danfruehauf:
> Sounds good to me. Yes the new workflow looks good @danfruehauf, nice work putting all that together!

I did the easy part. Talend is the hard part.

danfruehauf:
@xhoenner And this also needs merging. I'll merge the chef part.

PR #355, opened 2016-01-12
Fix error in conversion to trigger based processing

jonescc:


workspace/SOOP_TMV/code/routines/system/GeometryOperation_0.1.item
```diff
@@ -255,7 +297,7 @@ public class GeometryOperation {
      * 
      */
     public static Geometry GETGEOMETRYN(Geometry geom, int n) {
-	return geom.getGeometryN(n);
+    return geom.getGeometryN(n);
```
danfruehauf: Talend actually destroyed the indent here. But... Who cares.

PR #376, opened 2016-03-03
AUS PHYTO DB

lbesnard:
* add new columns, clean LQB.
 * needs new schema


xhoenner:
Ready to be reviewed (needs to be up urgently I believe). 

To deploy, objects in the `aus_phyto_db` schema need to be dropped.

@atkinsn you'll have to modify the existing metadata record to have a WFS layer in the distribution section called imos:aus_phyto_db_data.

mhidas:
I have only one main suggestion: Since `PROJECT_ID` is used in an `INNER JOIN` between the data and metadata tables, this column should have 1) a primary key constraint in the metadata table, and 2) a foreign key constraint in the data table. Otherwise, if any data happens to have an incorrect or null `PROJECT_ID`, it will just be silently excluded from the `aus_phyto_db_data` view.

The rest are mostly "cosmetic":
 * In the **metadata_update** subjob, I don't think the `tPostJob` and `tPostgresqlClose` components are doing anything (since there are no open connections there) and should be removed.
 * In the liquibase/SQL:
   * It might be nice to separate the `apd_data` and `apd_metadata` tables in separate changeSets.
   * Why not specify the full geometry column type (and even the check) when creating the table, instead of using an `ALTER TABLE` later? E.g. `geom geometry(Geometry,4326) CHECK (st_isvalid(geom))`
   * Consistent indentation would make the SQL a bit easier to read (especially in changeSet `create_aus_phyto_db_data_view`). Same in the `tPostgresqlInput_1` query in **data_harvest** subjob.

mhidas:
@smancini @atkinsn How urgent is this?

I haven't actually tested it, but @xhoenner says it works. If it needs to be deployed ASAP we can merge it. Otherwise I think it would be worth adding those constraints first (it wouldn't take too long).

smancini:
It will be good to finish this work by the end of the week.
@xhoenner How long will it take to make these changes?

xhoenner:
Thanks @mhidas for reviewing this PR, I've implemented all the changes you suggested except for the primary key/foreign key because some `PROJECT_ID` instances in the data table are not not present in the metadata table. To make sure all the data is available to users, I've changed the `INNER JOIN` to a `FULL JOIN`. Thanks for pointing that out Marty, I wouldn't have realised some PROJECT_IDs were missing in the metadata table otherwise

danfruehauf:
In a review, nothing escapes @mhidas's eyes. Really.

@xhoenner Let me know when you're ready to deploy that stuff and we'll do the schema drop etc.

mhidas:
Is that a problem that should be fixed in the source db at CSIRO before the final version is published?

xhoenner:
> Is that a problem that should be fixed in the source db at CSIRO before the final version is published?

That's a question for @atkinsn and @smancini. If you want to chat to Claire Davies, one of the projects that has data but no corresponding metadata is `PROJECT_ID` 1076.

atkinsn:
@xhoenner it does seem like it should be fixed. Can you draft something describing that, and I will contact Claire and see if she and Steve want/need to fix it.

xhoenner:
@atkinsn basically there are five project_ids (i.e. "1067","1068","1069","1070", "29") for which there are data available but no metadata (e.g. "PROJECT_DESCRIPTION", "SAMPLE_METHOD"). The opposite is also true (albeit less important): two project_ids (i.e. "736", "740") have metadata information but no data.

xhoenner:
All fixed now, please review and merge ASAP so that the harvester can run tonight as per @smancini's request. 

Could someone also review and merge https://github.com/aodn/geoserver-config/pull/330 so that we can test out the whole thing tomorrow?

mhidas:
I'll take a look. 

(This would be sooo much easier if the liquibase script could be in the repo as a separate xml file instead of squeezed in as in-line content within the subjob's ".item" file.)

mhidas:
Looks ok. Now that you have those constraints, you could change the `FULL JOIN` back to `INNER JOIN`, but it should make a difference.

PR #378, opened 2016-03-09
Add basic smoke tests for netcdf based talend components

jonescc:
Ensures components compile and produce the correct number of records for test data


danfruehauf:
@jonescc Perhaps think about how those can be triggered automatically from jenkins.

jonescc:
Unit tests talend project probably needs to be moved to the talend_components repo and built and run there on modification of a component.  Needs a bit of work

PR #379, opened 2016-03-15
FAIMMS - fix time coverage

lbesnard:
* issue with *._map having wrong value of time_start and time_end for a
   channel_id. Value was taken from last file added, and not from all
   channel. This is now fixed
 * also created an update query to fix already processed channels


lbesnard:
@xhoenner this should fix the issue in reporting

danfruehauf:
@lbesnard @xhoenner Any of you can trigger it with:
```
$ talend_liqui faimms-faimms
```

This will trigger the harvester liquibase migrations.

lbesnard:
@danfruehauf already done

lbesnard:
def a good thing to have created this function !

PR #381, opened 2016-03-16
ANMN NRS DAR YON - fix time coverage issue with *_map

lbesnard:


lbesnard:
@xhoenner this fixes similar issue as in FAIMMS

PR #384, opened 2016-03-17
SRS LJCO WWS

lbesnard:
* remove geom indexes which slowed down db
 * ADD instead geom column to deployments table and use this in views
 * fix view name hourlt to hourly


lbesnard:
linked to https://github.com/aodn/geoserver-config/pull/334

danfruehauf:
But wait, aren't the geom indexes required?

lbesnard:
on a different table. one geom per file now, instead of one geom per data point

PR #385, opened 2016-03-18
SRS OC LJCO WWS - RE drop index

lbesnard:
* retry to force the deletion of spatial indexes. tpostgresouput was
   forcing this. Replaced it with ipostgres fixed it, probably because
   it overwrote default options


danfruehauf:
@lbesnard Better works...

PR #386, opened 2016-03-18
SRS_ALTIMETRY trigger

danfruehauf:


danfruehauf:
Can be safely merged, as harvester is disabled until it'll get a pipeline.

PR #390, opened 2016-03-24
Reactivate all subjobs in AUS_PHYTO_DB for deployment to prod

xhoenner:
Can be deployed on Jenkins to prod, need to run on production only once. Existing objects in `aus_phyto_db` schema need to be dropped.


xhoenner:
Can someone please deploy to prod?

PR #392, opened 2016-04-11
AUV piplines

lbesnard:
* mod AUV_VIEWER_TRACK harvester to be pipeline proof
 * mod AUV harvester (netcdf only). Add metadata update component,
   fixing many things to be pipeline proof


bpasquer:
This one is good to go too @lbesnard

PR #394, opened 2016-04-18
SOOP-BA :correct bad url pointing to imos-t3

bpasquer:
@lbesnard  same issue as https://github.com/aodn/harvesters/pull/393


lbesnard:
:+1:

PR #395, opened 2016-04-18
SOOP ASF and SST - add investigator

lbesnard:
* add investigator to harvester, and changed from looking at vessel
   name to vessel code(less risk of error)
 * change was done on 2 harvesters


lbesnard:
@smancini FYI

bpasquer:
👍

PR #396, opened 2016-04-26
Modify context for metadata, deactivate all subjobs but MetadataUpdate

xhoenner:
for AUS_PHYTO_DB harvester. Data bags need to be amended, see config in workspace/AUS_PHYTO_DB/context/Metadata_0.1.item

Spatial resolution of 5 should be sufficient, @atkinsn. Please review and deploy ASAP.


lbesnard:
https://github.com/aodn/chef-private/pull/1825

PR #399, opened 2016-04-28
AODN IMAS FLUOROMETRY

xhoenner:
New harvester hence new schema needed and new set of data bags. For data bags config see configs in 
the following two files:
   * workspace/AODN_IMAS_FLUOROMETRY/context/source_0.1.item
   * workspace/AODN_IMAS_FLUOROMETRY/context/Metadata_0.1.item


lbesnard:
related to chef-private https://github.com/aodn/chef-private/pull/1831

PR #410, opened 2016-05-18
Add URL and size columns in _map table for AATAMS_SATTAG_NRT to support BODAAC

xhoenner:
Ready for review. Once this PR has been merged and the harvester deployed on 14-nsp, then the corresponding metadata record will have to be updated accordingly.


bpasquer:
Nice and simple, that's how we love PRs. Thanks @xhoenner

PR #419, opened 2016-06-07
SRS OC - add views and point to new file version

lbesnard:
* add views for par, dt and owtd aqua files
 * new files were uploaded with new hierarchy and filename. Changed the
   views to point to these


lbesnard:
easy PR. only changed views. Will have to be tested on RC with gogoduck to make sure aggregation can be done with new files

PR #420, opened 2016-06-08
ANMN NRS LONG TS - new harvester

lbesnard:
# NOT TO MERGE

Harvester for 
http://imos-data.s3-website-ap-southeast-2.amazonaws.com/?prefix=IMOS/ANMN/NRS/NRSMAI/aggregated_products/

Related PR's :
https://github.com/aodn/chef/pull/316
https://github.com/aodn/chef-private/pull/1922


lbesnard:
@mhidas happy to show you the views

mhidas:
@lbesnard I've added a copy of the liquibase script as a separate file. This makes it easier to review & comment.

workspace/ANMN_NRS_LONG_TS/code/liquibase/anmn_nrs_long_ts.changelog.xml
```diff
@@ -0,0 +1,129 @@
...
+           file_id integer UNIQUE NOT NULL,
+           time_coverage_start timestamp with time zone,
+           time_coverage_end  timestamp with time zone,
+           site_code character varying,
+           longitude double precision,
+           latitude double precision,
+           geospatial_vertical_min double precision,
+           geospatial_vertical_max double precision,
+           geospatial_lat_max double precision,
+           geospatial_lon_max double precision,
```
mhidas: I don't think we need `geospatial_lat_max` & `geospatial_lon_max` here, since they *should* be the same as the _min values (which are stored in `latitude` and `longitude`). However, as a just-in-case, it would be good to be consistent and create the geometry from the _min values too (currently it's coming from the _max values).

workspace/ANMN_NRS_LONG_TS/code/liquibase/anmn_nrs_long_ts.changelog.xml
```diff
@@ -0,0 +1,129 @@
...
+           measurement bigserial NOT NULL,
+           file_id integer NOT NULL,
+           "TIME" timestamp with time zone,
+           "LATITUDE" double precision,
+           "LONGITUDE" double precision,
+           "DEPTH" real,
+           "PSAL" real,
+           "TEMP" real,
+           CONSTRAINT measurements_pkey PRIMARY KEY (measurement),
+           CONSTRAINT fk_file_id FOREIGN KEY (file_id)
```
mhidas: Don't you need to give this constraint a different name? You already used `fk_file_id` above.

workspace/ANMN_NRS_LONG_TS/code/liquibase/anmn_nrs_long_ts.changelog.xml
```diff
@@ -0,0 +1,129 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+
+    <changeSet author="talend" id="cr_deployments-1">
+       <sql>
+         CREATE TABLE deployments
```
mhidas: These are not deployments. `timeseries` would be a better name for this table.

workspace/ANMN_NRS_LONG_TS/code/liquibase/anmn_nrs_long_ts.changelog.xml
```diff
@@ -0,0 +1,129 @@
...
+        CREATE VIEW anmn_nrs_long_ts_map AS
+          SELECT 
+            d.file_id,
+            f.url,
+            f.size,
+            site_code,
+            time_coverage_start AT TIME ZONE 'UTC' AS time_coverage_start,
+            time_coverage_end  AT TIME ZONE 'UTC' AS time_coverage_end, 
+            geospatial_lon_max,
+            geospatial_lat_max,
```
mhidas: Use `latitude` and `longitude` here instead of the _max columns.

workspace/ANMN_NRS_LONG_TS/code/liquibase/anmn_nrs_long_ts.changelog.xml
```diff
@@ -0,0 +1,129 @@
...
+    </changeSet>
+
+
+    <changeSet author="talend" id="anmn_nrs_long_ts_data-1" runOnChange="true">
+      <sql>
+        DROP VIEW IF EXISTS anmn_nrs_long_ts_interpolated_data; 
+        CREATE VIEW anmn_nrs_long_ts_interpolated_data AS
+          SELECT
+            measurements.file_id,
+            a.site_code,
```
mhidas: It would be nice to consistently use either full table names (measurements) or aliases (a, f), rather than a mix of both.

lbesnard:
@mhidas  amended all comments. wil squash commits if you're happy

mhidas:
Thanks @lbesnard . The idea with the separate copy of the liquibase xml file is that it should be updated to match what is in the code. It's annoying that you can't just tell Talend to read the liquibase script directly from this file, but cutting and pasting it only takes 5 seconds. (I usually find it easier to edit the script in an external text editor anyway, rather than in Talend directly.) Then a reviewer can easily see on GitHub what has changed, without having to check out and open the project in Talend. It's just a convenience thing.

workspace/ANMN_NRS_LONG_TS/code/liquibase/anmn_nrs_long_ts.changelog.xml
mhidas: What I meant by this is that you should make sure the same value goes into `latitude`/`longitude` and `geom`, even if the assumption that lat_min==lat_max and lon_min==lon_max turns out to be incorrect. Currently this is still not the case (you're averaging _min & _max to get geom, which is fine, but then only putting the _min values into `latitude`/`longitude`)

lbesnard:
@mhidas This is rebased and ready to be merged

PR #421, opened 2016-06-14
Liquibase code to add imos_b column to installation_summary table and…

xhoenner:
… update corresponding subjob.

Linked to https://github.com/aodn/geoserver-config/pull/362, so that users can visualise on Step 2 the location of all IMOS receivers (was requested by Tim a few weeks ago).


bpasquer:
The imos_b variable is only necessary for layers visualization, not for reporting , cause i couldn't find a change in the Populate_reporting view job. Right?

xhoenner:
> The imos_b variable is only necessary for layers visualization, not for reporting

Correct. Thanks @bpasquer for reviewing this PR!

PR #424, opened 2016-06-21
Get rid of AATAMS_SATTAG_DM materialised view

xhoenner:
* Change Liquibase to get rid of materialised view and add columns to summary table instead
  * Add SQL query to update summary table

Addresses https://github.com/aodn/internal-discussions/issues/228 @lwgordonimos


julian1:
@xhoenner How does this work to address https://github.com/aodn/internal-discussions/issues/228 . Does it reduce the amount of data needing to be copied?

lwgordonimos:
Looks like it updates the summary table on each run, rather than requiring a whole refresh of the view each time.

(ripped this out and formatted/decoded it)
```sql
UPDATE summary SET lat = b.lat, lon = b.lon, geom = b.geom 
FROM (
WITH a AS (
         SELECT summary_id,
        m.device_id,
        timestamp_start,
        CASE WHEN st_npoints(me.geom) = 1 THEN st_y(me.geom) ELSE st_y(st_pointn(st_geometryn(me.geom, 1), 1)) END AS lat,
                CASE WHEN st_npoints(me.geom) = 1 THEN st_x(me.geom) ELSE st_x(st_pointn(st_geometryn(me.geom, 1), 1)) END AS lon,
                CASE WHEN st_npoints(me.geom) = 1 THEN me.geom ELSE st_pointn(st_geometryn(me.geom, 1), 1) END AS geom,
        row_number() OVER (PARTITION BY m.summary_id ORDER BY abs(date_part('epoch'::text, m.timestamp_start - me.timestamp))) AS rn
           FROM summary m, aatams_sattag_dm_profile_map me
          WHERE (m.timestamp_start - me.timestamp) <= '5 days'::interval AND (m.timestamp_start - me.timestamp) >= '-5 days'::interval AND me.device_id = m.device_id
          ORDER BY m.summary_id, device_id, timezone('UTC'::text, m.timestamp_start)
        )
 SELECT summary_id, device_id, timestamp_start, lat, lon, geom FROM a WHERE a.rn = 1
UNION ALL
 SELECT summary_id,
    device_id,
    timestamp_start,
    NULL::double precision AS lat,
    NULL::double precision AS lon,
    NULL::geometry AS geom
   FROM summary WHERE NOT (summary_id IN (SELECT a.summary_id FROM a WHERE a.rn = 1))) AS b
   WHERE summary.summary_id = b.summary_id;
```

julian1:
When the mat-view gets "refreshed" it should already be pretty efficient in terms of only updating what it has too.

@xhoenner ?

xhoenner:
> Looks like it updates the summary table on each run, rather than requiring a whole refresh of the view each time.

It does indeed. Originally I was thinking to only update the records for the files that are being processed but that would have been an issue for files that have already been harvested. I ran the harvester using a backup of the dbprod schema, the update part outlined above took less than 5 minutes to run on my VM.

PR #425, opened 2016-06-23
Add a generic timestep harvester for CARS/OA Reconstruction

jonescc:
Takes schema details and time variable as input.  Could be made more generic by including collection as a parameter.


lbesnard:
@jonescc
I only have 2 minor comments :
-we generally respect the case of netcdf variables in the column names of the DB. For what this convention is worth, it means that "time" should be "TIME" in both the table and view
-the talend recycle bin can be emptied

awesome otherwise

jonescc:
@lbesnard - changes made

PR #427, opened 2016-07-04
Add collection parameter

jonescc:
@lbesnard, I've added a collection parameter to the generic timestep harvester so that we can use a single schema for collections harvested by this harvester where suitable.


lbesnard:
thanks @jonescc . 

I'm just a bit confused how setCollectionId gets the ```collection.id``` value rather than ```collection``` but I believe it works

PR #430, opened 2016-07-12
Remove filter custom from harvesters

lbesnard:
* the modified harvesters, were using the custom query option on
   iNewOrModifiedFileList to identify which files are new or modified, but the
   query used is based on pre-manifest harvesting logic and is considering the
   file to have been added rather than deleted causing it to fallover
   when trying to read the file.
   -> this is now modfified to use the normal filtering rather than
   custom query
 * the SRS OC harvester was modified not to use the custom query anymore
   since some work on CF Date format was done. The harvester is simpler,
 * The harvest_type is updated to a new value in the file_harvest table
   so there is no need for different subjobs per product


jonescc:
@lbesnard - you've used the netcdf variable name to talend column name mapping option on the advanced tab mapping "time" to "time".  The column name should not be surrounded by quotes as its not something that can be substituted with an expression  (the talend component includes quotes in the default value but its not required so this is misleading).  This will result in the mapping not working, but I hoipe you can see from my description of what this option is used for, that you actually don't need to use this option anyway.  There is no need to map column "time" to variable "time" - this is the default option.  Just remove this row from the mapping table.

lbesnard:
Thanks Craig. They all work now on my PO

PR #431, opened 2016-07-12
ANMN metadata harvester updates

mhidas:
Various fixes and updates in preparation to deploy this for ABOS as well:
* determine whether a file is real-time using a more general regex
* define getglobalattribute functions before setting up schema
* add new data category "aggregated_products"
* update checker signature attributes


workspace/ANMN_METADATA/code/sql/file_metadata.update.sql
```diff
@@ -39,7 +39,7 @@ UPDATE file_metadata m
     data_category = coalesce(substring(url, '/(Temperature|(CTD|Biogeochem)_(timeseries|profiles)|Velocity|Wave|CO2|Meteorology|Surface_[^/]+|Sub-surface_[^/]+|Sediment_traps)/'), substring(url, '/(Pulse|FluxPulse)/')),
     variables = v.variables,
     standard_names = s.standard_names,
-    realtime = (f.url ~ 'realtime|REAL_TIME'),
+    realtime = (f.url ~* 'real[-_]?time'),
```
ggalibert: It's a small detail but I think **when possible** it is better to use faster SQL standard operators like LIKE rather than postgresql specific support for posix regular expressions. So line 7 could be replaced by:
```SQL
value NOT LIKE '%flag'
```
Line 12:
```SQL
name NOT LIKE '%_quality_control'
```
However line 42 just above might not be able to translate into a simple SQL standard query and posix regexp would be a better way to go.

See https://www.postgresql.org/docs/9.4/static/functions-matching.html

workspace/ANMN_METADATA/code/sql/file_metadata.update.sql
```diff
@@ -36,7 +36,8 @@ UPDATE file_metadata m
     time_deployment_start = getglobalattributeastimestamp(f.id, 'time_deployment_start'),
     time_deployment_end = getglobalattributeastimestamp(f.id, 'time_deployment_end'),
     geom = st_geomfromtext('POINT(' || longitude::text || ' ' || latitude::text || ')', 4326),
-    data_category = coalesce(substring(url, '/(Temperature|(CTD|Biogeochem)_(timeseries|profiles)|Velocity|Wave|CO2|Meteorology|Surface_[^/]+|Sub-surface_[^/]+|Sediment_traps)/'), substring(url, '/(Pulse|FluxPulse)/')),
+    data_category = coalesce(substring(url, '/(Temperature|(CTD|Biogeochem)_(timeseries|profiles)|Velocity|Wave|CO2|Meteorology|Surface_[^/]+|Sub-surface_[^/]+|Sediment_traps|aggregated_products)/'),
+                             substring(url, '/(Pulse|FluxPulse)/')),
```
ggalibert: Is it intended that there is not a category for every data file? Wind and water level stuff don't seem to belong to any of the above categories.

mhidas: As far as I can tell this will get a category for every ANMN & ABOS file except for the NRS Darwin & Yongala real-time files, which are organised very differently and are therefore in the too-hard basket for now. We can re-visit that in a future PR.

PR #432, opened 2016-07-14
GENERIC TIMESTEP - pwd context fix

lbesnard:
* Destination_Password was of type Password rather than String.
   Conflicted with databags


lbesnard:
@jonescc FYI

PR #435, opened 2016-07-18
ANMN_ACOUSTICS: harvest deployment metadata from csv file

mhidas:
A simple harvester to read metadata from a csv file and insert/update into the `acoustic_deployments` table.


mhidas:
@xhoenner @ggalibert This one is ready to merge. Works on PO Box, so no need to test. Just have a quick look.

PR #439, opened 2016-07-28
Re-activated and modified subjob to update metadata table + modified …

xhoenner:
…_data view to have presence_absence field as boolean.

Amendments requested by Claire Davies at CSIRO. The sooner it's merged, the sooner it can be deployed @smancini, @atkinsn.


mhidas:
@xhoenner Looks ok, just two minor comments:

1) In the liquibase:
```sql
CASE WHEN data."PRESENCE_ABSENCE" = 1 THEN TRUE ELSE FALSE END AS "PRESENCE_ABSENCE"
```
is exactly equivalent to
```sql
data."PRESENCE_ABSENCE" = 1 AS "PRESENCE_ABSENCE"
```
so there's no need for a `CASE` expression.

2) In the main job `AUS_PHYTO_DB_harvester`, you have several subjobs linked by `OnSubjobOk` connections, but most of them are disabled. Does this run as expected? e.g.  does `metadata_harvest` only run if `iPostgresqlDbUpdate_1` succeeds, and does it always run before `metadata_update`?

PR #443, opened 2016-08-09
WAMSI harvester

xhoenner:
New Talend harvester to harvest data from a single shapefile into the database. This shapefile will soon be uploaded to data.aodn.org.au so testing will be possible then. @smancini could you please upload the corresponding shapefile and its associated files to data.aodn for me and create the directory structure as highlighted in Luke's email? 

Data bags will be needed, new schema and new login role.


xhoenner:
Ready for review and testing as the data is now available on 10-aws-syd in /mnt/imos-data/WAMSI/WAMSI1/Node3/3.4/3.4.2/Ningaloo_Inshore_Offshore_Sediment_Data.

mhidas:
@xhoenner Looks good to me. 

The only minor comment I'd make, is that I don't think there's any benefit to specifying an arbitrary maximum length to the "character varying"-type columns in the table. You could just use "character varying" without a number afterwards, or simply "text" (see [PostgreSQL docs](https://www.postgresql.org/docs/9.4/static/datatype-character.html) for details)

mhidas:
And I think the `include` pattern should be ".*\.shp" (it's a regular expression, so '.' matches anything, to match only the dot itself, you need to escape it with '\')

mhidas:
I haven't done all the data bags, but I've made a start by creating the db role and schema for the PO Box: https://github.com/aodn/chef/compare/aodn_wamsi_sediment_harvester?expand=1

xhoenner:
thanks @mhidas for reviewing this PR.

> I don't think there's any benefit to specifying an arbitrary maximum length to the "character varying"-type columns in the table

Neither do I but Talend yelled at me when I tried to leave the numbers blank in the table schema (in iPostgresqlOutput).

> And I think the `include` pattern should be `".*\.shp"`

sounds good, I'm not an expert in regex so I'm trusting you on that. The harvester doesn't need to be changed though as the `include` is specified in the data bags.

mhidas:
> I don't think there's any benefit to specifying an arbitrary maximum length to the "character varying"-type columns in the table 

Sorry, I forgot to specify, I was talking about the table creation SQL in the liquibase. But I don't think it's a big deal, unless the input file ever ends up having longer strings in any of those columns.

mhidas:
@xhoenner This should be deployed on 14-nec-hob whenever it provisions successfully, and on prod at 5pm tonight.

mhidas:
Note that the source file is not actually accessible on the rc machine.

xhoenner:
Thanks @mhidas, do you want me to copy the source file in my home directory on 14-nec-hob so that the harvester can be tested on this machine before being deployed on prod?

xhoenner:
@atkinsn could you please push the corresponding metadata record to catalogue-portal?

PR #444, opened 2016-08-30
  Soop co2 update and improvements

bpasquer:
- restructure the harvester to facilitate the addition of the iUpadeMetadata component
 - standardise name for time_start time_end in map view (becomes time_coverage_start,_end)


bpasquer:
Although it could look like a major update, it is just about addig the iUpdate Metadata component(!) and changing the name of two variables in the map_view.

bpasquer:
@xhoenner  the change of name for time_start and time_end (as i mentioned to you last week) is part of this PR

PR #445, opened 2016-08-31
SOOP-SST: Rename time variable columns in soop_sst_nrt and soop_sst_d…

bpasquer:
…m map table

 - add liquibase command
 - edit subjobs accordingly


bpasquer:
Please do not merge + deploy before the monthly reporting has ran

PR #448, opened 2016-09-02
 SOOP BA update:

bpasquer:
- rename time variables
 - change context destination server to point to vagrant
 - add UpdateMetadata component


bpasquer:
Similar to https://github.com/aodn/harvesters/pull/445

PR #449, opened 2016-09-02
 SOOP_TMV update:

bpasquer:
- rename time variables in liquibase and update subjobs
 - edited destination server for testing purposes

 edited destination server for testing purposes


bpasquer:
one more simlar to https://github.com/aodn/harvesters/pull/445

xhoenner:
Note that all PRs with the 'time_coverage' change can now be deployed since Jac has produced the data reports. I wouldn't be surprised though that the reporting views have to be dropped beforehand for the harvester to succeed so let me know and I'll do that for you.

bpasquer:
The sooner the schemas can be change the better.
thanks @xhoenner

PR #451, opened 2016-09-02
SOOP_TMV_NRT: rename time variables in liquibase and update subjobs

bpasquer:


bpasquer:
similar to https://github.com/aodn/harvesters/pull/445

PR #460, opened 2016-09-14
faimms - fix LQB wrong checksum

lbesnard:


lbesnard:
impossible to review. maybe due to a empty space

mhidas:
> impossible to review

You can make it quite easy to review liquibase changes if you make a copy of the liquibase xml file in the harvester's workspace, e.g. under `workspace/FAIMMS/code/liquibase/faims.liquibase.xml`, then always edit that first, then copy-paste it into Talend. It's a bit annoying to have to remember doing this just to keep it in sync, but makes the changes much easier to review.

@jonescc  did we ever figure out why we can't just point `iPostgresqlDbUpdate` to a separate xml file instead of having  the liquibase script inline?

jonescc:
When I looked at this, I thought the best place for the xsl would be in the project e.g. in a resources directory in the talend project workspace.  But talend only includes directories and items it knows about in the packaged zip file - so I would have had to modify the talend build process to include it.   At the time I thought this was too much work.  

There are probably ways to do this now - we do run a custom talend build process - so we could include resources in a resource directory in a jar for example and add an option to the iPostgresqlDBUpdate component to look for file on the class path.  Its would have also been possible to source the xsl file from somewhere else - e.g. a web accessible resource - but this seemed to be overcomplicating things at the time.  

Perhaps raise this as a possible backlog item?

mhidas:
Thanks @jonescc . Right... I'd forgotten that deployment was the problem. I added a backlog item: https://github.com/aodn/backlog/issues/437

lbesnard:
> You can make it quite easy to review liquibase changes if you make a copy of the liquibase xml file in the harvester's workspace, e.g. under workspace/FAIMMS/code/liquibase/faims.liquibase.xml, then always edit that first, then copy-paste it into Talend. It's a bit annoying to have to remember doing this just to keep it in sync, but makes the changes much easier to review.

in this case, the diff is pointless. There is no change. but different checksum. maybe a space somewhere

PR #468, opened 2016-10-13
SRS ALTIMETRY - CF & IMOS 1.3 compliance

lbesnard:
* modified harverster for new files complying with CF and IMOS 1.3
   version
 * use generic time to harverst time
 * rename all DEPTH column to NOMINAL_DEPTH
 * get instrument and site site_code from gatts, and modified harverster
   accordingly


bpasquer:
I am just wondering why you begin the  job updateMetadata with a TPostJob. There is no risk of having an exception/ error , and nothing to be cleaned up.

lbesnard:
@bpasquer I don't think it matters that much, as it closes a connection if it stays open. But I removed it anyway

PR #474, opened 2016-10-31
Create AODN_NSW_OEH_bathymetry harvester

mhidas:
This is a new harvester for multi-beam bathymetry data from NSW OEH


mhidas:
I've renamed the harvester to remove the "bathymetry" from the name, as there will be other products from NSW OEH later. PR is #476 replaces this one.

PR #482, opened 2016-12-12
Remove talend component tests

jonescc:
Moved to talend-components repo where they belong.


julian1:
Github view on this is too slow to load, so merging as-is.

PR #485, opened 2016-12-18
Use axis type column mapping to map time variable

jonescc:
Automatically detects time variable rather than requiring it to be specified


jonescc:
@lbesnard

PR #490, opened 2016-12-22
Revert " Update url and harvester cleanup:"

lbesnard:
Reverts aodn/harvesters#484


lbesnard:
merged with @bpasquer

PR #496, opened 2017-02-16
ARGO - Major update

lbesnard:
* Processing ALL floats regardless of them being AUS or not
 * Add variables re Merged-argo files
 * add columns for for wms/wfs to tell if a profile is a Bio, core or
   merged. As well as Delayed or Realtime. Can be used by geoserver
   filters
 * fix bug in platform_metadata table where some files had a new
   dimension N_POSITIONING_SYSTEM > 1 which created an issue with a
   unique pkey


lbesnard:
will have to be tested on RC since we can't test geoserver filters on PO box

PR #499, opened 2017-02-24
SOOP XBT DM - fix wms to read callsign from Platform_code att as well

lbesnard:


lbesnard:
FYI @smancini . we will have to repush all the files

PR #500, opened 2017-03-02
Argo  improvements

lbesnard:
* rename main data view to be more consistent
 * added few columns in main data view to be used in child views
 * created few children data views for multi wfs (to be tested)


lbesnard:
modification of the views. This will break geoserver rc. but at least can be tested on PO box. can be merged and deployed to RC only

Issue #504, opened 2017-03-10
ANFOG_DM harvester failing on 10aws


bpasquer:
ANFOG_DM harvester is failing on 10aws. 
The error is the same as for AODN_RAN_CTD https://github.com/aodn/harvesters/issues/501

See the following log for full details IMOS_ANFOG_BCEOPSTUV_20170109T100117Z_SG540_FV01_timeseries_END-20170131T204838Z.nc.EbYY92.20170310-112353.log

> Going to process a total of '1' files
Processing slice with '1' files
Preparing file '/mnt/ebs/tmp/IMOS_ANFOG_BCEOPSTUV_20160228T225753Z_SL239_FV01_timeseries_END-20160330T073251Z.nc.xpGyCy,IMOS/ANFOG/slocum_glider/Portland20160229/IMOS_ANFOG_BCEOPSTUV_20160228T225753Z_SL239_FV01_t
imeseries_END-20160330T073251Z.nc' in temp dir '/tmp/d20170307-27783-1tjbpem'
ADD '/mnt/ebs/tmp/IMOS_ANFOG_BCEOPSTUV_20160228T225753Z_SL239_FV01_timeseries_END-20160330T073251Z.nc.xpGyCy' => 'IMOS/ANFOG/slocum_glider/Portland20160229/IMOS_ANFOG_BCEOPSTUV_20160228T225753Z_SL239_FV01_timeser
ies_END-20160330T073251Z.nc'
'anfog_dm-anfog_dm': ["IMOS/ANFOG/slocum_glider/Portland20160229/IMOS_ANFOG_BCEOPSTUV_20160228T225753Z_SL239_FV01_timeseries_END-20160330T073251Z.nc"]
Executing '/mnt/ebs/talend/jobs/anfog_dm-anfog_dm/java/ANFOG_DM_harvester/ANFOG_DM_harvester_run.sh --context_param paramFile="/mnt/ebs/talend/jobs/anfog_dm-anfog_dm/etc/anfog_dm-anfog_dm.conf" --context_param ba
se=/tmp/d20170307-27783-1tjbpem --context_param fileList=/tmp/harvester_file_list20170307-27783-s6auuw --context_param logDir=/tmp/d20170307-27783-phi06f' for '["IMOS/ANFOG/slocum_glider/Portland20160229/IMOS_ANF
OG_BCEOPSTUV_20160228T225753Z_SL239_FV01_timeseries_END-20160330T073251Z.nc"]'
<...>
[...]

Issue #501, opened 2017-03-03
AODN_RAN_CTD harvester failing in RC and prod


bpasquer:
When manually triggered in RC and prod the AODN_RAN_CTD harvester( no pipeline for this collection) outputs the following error:

"_Exception in thread "main" java.lang.Error: java.lang.Error: java.lang.Error: java.lang.Error: java.lang.Error: java.lang.Error: java.lang.Error: java.lang.AbstractMethodError: org.slf4j.impl.JDK14LoggerAdapter.log(Lorg/slf4j/Marker;Ljava/lang/String;ILjava/lang/String;[Ljava/lang/Object;Ljava/lang/Throwable;)V
        at aodn_ran_ctd.aodn_ran_ctd_harvester_0_1.AODN_RAN_CTD_harvester.tRunJob_2Process(AODN_RAN_CTD_harvester.java:1199)
        at aodn_ran_ctd.aodn_ran_ctd_harvester_0_1.AODN_RAN_CTD_harvester.runJobInTOS(AODN_RAN_CTD_harvester.java:3262)
        at aodn_ran_ctd.aodn_ran_ctd_harvester_0_1.AODN_RAN_CTD_harvester.main(AODN_RAN_CTD_harvester.java:3070)_"

The harvester AODN_RAN_CTD cannot be tested on the PObox since it's harvesting from thredds server. But it ran successfully on my local machine.  Some help to understand the meaning of the error message would be very much appreciated.

jonescc:
Has been rebuilt and promoted to rc.  Can this be tested there before promoting to RC?

bpasquer:
yes, i've just triggered talend on RC

bpasquer:
AODN_RAN_CTD harvester successfully ran on RC .

bpasquer:
Can be promoted to prod. Thanks

jonescc:
Promoted - will be deployed after 5pm this afternoon.

PR #505, opened 2017-03-14
Remove reference to old version of netcdf library - talend can pick i…

jonescc:
…t up from installed components

Most likely cause for https://github.com/aodn/harvesters/issues/504 and https://github.com/aodn/harvesters/issues/501.


jonescc:
@bpasquer - can you review this?  If we build anfog_rt again without this fix, we'll end up with issues like #504 and #501 again.

bpasquer:
Hold on Craig, are you fixing ANFOG_DM or ANFOG_RT or both?

jonescc:
ANFOG_RT includes an old version of the netcdf java library.  This was recently built in jenkins causing the old version of the netcdf java library to be copied to the harvester build library.  ANFOG_DM and AODN_RAN_CTD were then built in jenkins and the old version of the library was copied from the harvester build library along with the required version.   This then caused #504 and #501.  So not fixing ANFOG_DM but stopping ANFOG_RT from causing similar issue in future.

bpasquer:
ok, i see

bpasquer:
It solved the issue on the PO box.  Could you please deploy on RC @lbesnard

lbesnard:
will be deployed on RC at lunch

PR #508, opened 2017-03-22
ARGO - add nitrate and ph data

lbesnard:


bpasquer:
@lbesnard 
Create_Tabe_views, IPostgresqlDBUpdate5, in the argo_profile_nitrate_data view: You've left quotes around the variable position. It certainly not impacting on the processing but, is there a reason why you've put quotes?

PR #509, opened 2017-03-27
 Harvester update for processing of new files:

bpasquer:
- add new variable to measurements table and _data view in liquibase
- update HarvestTMV subjobs


bpasquer:
Fixes issue with missing geometry and timestamp in  soop_tmv_ map view when processing latest version of DM soop TMV files.

PR #511, opened 2017-03-30
Removed old netcdf jars, replaced hardcoded path with context variables

bpasquer:


bpasquer:
Minor changes, mostly cleaning up hardcoded path, and removing unwanted files

PR #513, opened 2017-04-11
 Issue fixing and cleaning up of old indexed files

bpasquer:
- Run clean up of metoc files in data tables using liquibase query in DataHarvest.
- Fix issue with computation of buoy specific info in HarvestData


bpasquer:
Could you please deploy to RC @lbesnard

PR #521, opened 2017-06-05
New harvester for AUS_CHLA_DB

xhoenner:
Will require new db schema and login role. Harvester only needs to run once since this dataset is to support a publication, therefore no pipeline required, just need to be able to trigger the harvester manually.


bpasquer:
@xhoenner  2 minor changes.
- Job metadat_harvest, Tmap component Untick "nullable"  for PROJECT_ID. A primary key is not nullable.
- job data_harvest, tFileFetch component : tick "Die on error"

xhoenner:
Thanks @bpasquer, changes made in latest commit.

PR #525, opened 2017-06-20
SOOP ASF MT - index ISAR QC file but not harversting in schema

lbesnard:


workspace/SOOP_ASF_MT/process/SOOP_ASF_MT/SOOP_ASF_MT_harvester_0.1.item
```diff
@@ -165,9 +165,9 @@
     <elementParameter field="TEXT" name="PASS" value="context.Destination_Password"/>
     <elementParameter field="TEXT" name="JOBNAME" value="&quot;index&quot;"/>
     <elementParameter field="TEXT" name="HARVEST_TYPE" value="&quot;data&quot;"/>
-    <elementParameter field="RADIO" name="USE_DEFAULT" value="true"/>
-    <elementParameter field="RADIO" name="USE_FILTER" value="false"/>
-    <elementParameter field="TEXT" name="FILTER" value="&quot;%&quot;"/>
+    <elementParameter field="RADIO" name="USE_DEFAULT" value="false"/>
+    <elementParameter field="RADIO" name="USE_FILTER" value="true"/>
+    <elementParameter field="TEXT" name="FILTER" value="&quot;%FV01_C%&quot;"/>
```
mhidas: Are you sure the "ISAR QC" file names will never have a field starting with 'C' after the FV01? If not, the pattern could be "%FV01_C-2%"?

lbesnard: well it should not be as the files should not go beyond the watch.d regex, and secondly files should not be named this way. but yeap can add an extra layer of safety

lbesnard:
done

PR #527, opened 2017-07-05
 Update for pipeline deployment:

bpasquer:
- Change name of harvester to harmonize with schema name (best parctise for deployment)
 - Update context metadata_file to point to S3


bpasquer:
Could you please check this out @xhoenner

PR #528, opened 2017-07-07
 AATAMS ACCOUTSIC harvester update for implementation in pipeline

bpasquer:
- change name of main job. Now same as DB schemai
 - update context metadat_file. Context should point to S3.


bpasquer:
@xhoenner this is a new PR . If you have same issue as before please follow the step i suggested (delete hidden files; reload components)

PR #548, opened 2017-09-13
Remove old version of netcdf library added to tsunami buoys harvester

jonescc:
TOS copied these components to the components directory at some point.  They are not required and out-of-date.   Components should be sourced from the jenkins components directory which contains the required versions.


jonescc:
Suspected cause of https://github.com/aodn/issues/issues/97

PR #549, opened 2017-09-13
Remove unnecessary netcdfAll-4.2 import

jonescc:
None of these harvesters need this import statement - its included by iNetCDFInput which can have its version changed separately.  

Using routine library imports is problematic as they require harvesters using them to be updated on a version change.   I think it would be better to add a component which which can be updated independently.  Can add to iIncludeSdiComponents if its ever required.


jonescc:
Contributed to aodn/issues#97.   Will also stop TOS complaining about netcdfAll-4.2 being missing when building/running these harvesters

PR #554, opened 2017-10-19
FAIMMS - fix harvest of data previously done with timestamp without t…

lbesnard:
…imestamp information


lbesnard:
@smancini @anguss00 As we talked about it, it would be great to test this on RC first with a brand new schema. 
```sql
select drop_objects_in_schema('faimms');
```

ggalibert:
BTW, a timestamp is basically a date/time. I hope what you mean is that now time zone is taken into account.

lbesnard:
basically, the time value in the NetCDF file was not harvested properly. The TIME column definition was TIME WITHOUT TIME ZONE. And the Time zone from where is sitting the DB would be added to the TIME value such as
2016-06-01 00:00:00 -> 2016-06-01 10:00:00+00  which was completely wrong.

Changed the column definition to TIMESTAMP WITH TIMEZONE as mentioned here
https://docs.google.com/document/d/1-E05wDMvI_Yy2Zz9m3D8DCWUjjkhPDJFm_jr9TvbUgc/edit?pli=1#

Unfortunately, all the data need to be reharvested

ggalibert:
Well done for resurrecting this document!

If you are using a view for your WFS/WMS, don't forget to explicetely request a date/time in UTC too:
```SQL
SELECT
m.ts_id AS timeseries_id,
...
m."TIME" AT TIME ZONE 'UTC' AS "TIME",
...
FROM  measurement m
```

bpasquer:
The attributes time_coverage_start/end should also get fixed . 

See in : subjob noqaqc_table, table timeseries_features_deployments_varname 
and
subjob create_views,  view faimms_timeseries_map

lbesnard:
@bpasquer  it looks like it s doing what we want

PR #555, opened 2017-10-20
SRS OC SOOP RAD - handle CF time

lbesnard:
* change the harvester to read CF time value the same way as the
   generic timestep harvester


lbesnard:
This is done to process new data

bpasquer:
Ready for testing on RC

PR #559, opened 2017-10-25
New Harvester for AODN_WAVE_NRT

bpasquer:


mhidas:
Looks good, but I do have a few suggestions:

* Since you're using the `wmo_id` as the key when inserting rows into `aodn_wave_nrt_timeseries_map` table, that db column should also be created with the properties `NOT NULL` and `UNIQUE`. This is assuming the wmo_id values in that table are supposed to be unique and always defined. (Actually in that case you could even make `wmo_id` the primary key in that table.)

* In the `measurements` table, the `WMO_ID` column is used as a foreign key to join with `aodn_wave_nrt_timeseries_map`, so you should add a `FOREIGN KEY` constraint.

* In the component you've labelled "Update temporal extent of new and existing sites", I think the SQL would be much simpler if you just took the range of timestamps in the `measurements` table (which is supposed to contain *all* the data), rather than working form the temporary list of new data in `nrt_data`. You could just go ahead and update all rows, without worrying about what was already there.

* The iUpdateSpatialExtent component is disabled. Also, the Geonetwork URL is hard-coded to "http://localhost:8080/geonetwork" instead of using the relevant context parameter.

* You could simplify the liquibase a little by specifying all the properties of the `geom` column directly in the table definition (`geom geometry(Geometry,4326) CHECK (st_isvalid(geom)),`) instead of in a separate `ALTER TABLE` statement)

bpasquer:
After more testing, found out that the initial version was flawed. I made major changes with the addition of a table "deployments" that is updated with nrt data and the use of a deploy_no as a key.
Come chat to me if anything in the approach is  unclear

bpasquer:
@mhidas

mhidas:
That's better. A few more comments though...

### Subjob `Harvest_csv_metadata`

* The component `tMap_1` doesn't seem to do anything, so could be removed?


### Subjob `Harvest_NRT_data`

* In component `tConvertType_1` you drop several of the columns from the schema, including what look like data to me. Is that intended? Are there columns in the input data that you don't want to harvest? You also drop the Geometry column. Could this not be directly stored in the database, rather than re-creating it from the lat/lon columns?

* If each mesurement uniquely belongs to a deployment, and each deployment has a uniuqe `wmo_id`, then you should not need to repeat the `wmo_id` column in the `measurements` table. You just get it from the `deployments` table when you do the join!

* Table `deployments` should have a foreign key constraint for `wmo_id`.

* In `tPostgresqlRow_4` I don't see why you need two levels of sub-select. Wouldn't this do the same thing?
```sql
UPDATE nrt_data
SET deploy_no = (
  SELECT deploy_no
  FROM deployments
  WHERE nrt_data.wmo_id = nrt_data.\"WMO_ID\" AND
        nrt_data.longitude = nrt_data.\"LONGITUDE\" AND
        nrt_data.latitude = nrt_data.\"LATITUDE\"
  GROUP BY deploy_no
);
```

* In `tPostgresqlInput_1` the section 
```sql
WITH ...
nrt AS (
  SELECT *
  FROM nrt_data
)
```
does nothing. If you want to use a shorter name for the table you can just say ```FROM nrt_data nrt ...```

bpasquer:
I have adressedall you points .
As for the parameters I dropped from the WFS: they are not populated in the WFS and that have not been defined in the metadata record, so decided to leave them out.
 As for the field geom : it provides the SRS and the point location(lon,lat), so it is not a Geometry .

mhidas:
:+1: Nice work!

mhidas:
@bpasquer  Is this good to go into RC?

bpasquer:
YES please deploy

mhidas:
Ok, the harvester is built and will be deployed to RC once https://github.com/aodn/chef-private/pull/2541 is merged. Just waiting for @lbesnard to tweak the data bags a bit.

PR #560, opened 2017-10-31
FAIMMS - fix bugs - improve speed

lbesnard:
* Improve speed of queries by adding more specific where clauses
 * fix file not found issue in netcdf components


bpasquer:
Good for testing on RC

PR #571, opened 2017-11-21
ANMN_METADATA: update all timestamps in _map views to time zone UTC

mhidas:
Geoserver assumes all timestamps are in UTC, so we need to explicitly set this for every timestamp in all views used to create layers. See https://github.com/aodn/content/issues/297#issuecomment-345916918


mhidas:
This is a quick one to review. All changes can be seen in the [changelog.xml](https://github.com/aodn/harvesters/pull/571/files#diff-6b8d31234384ef8b852b0a9099298376)

PR #585, opened 2017-12-19
New harvester for CSV header schema "parameters_mapping"

xhoenner:
Related to https://github.com/aodn/chef-private/pull/2627. Original schema needs to be dropped first. 

Data bags need to be created.


mhidas:
The harvester is nice and simple and looks ok to me. I just have a couple of comments on the db table definitions:
* Table `unit_view` should have a primary key constraint (on `id` column I assume).
* Table `parameters_mapping` references `unit_view`, so should have a foreign key constraint.
* Isn't it a bit confusing to name a table `unit_view`? Why not just `units`?
* When defining text columns in the database, there's no need to put arbitrary length limits. Just use `character varying`, or `text` (which is the same thing).

xhoenner:
All done @mhidas

bpasquer:
Nothing major, just a few comments to make it cleaner
* You could add a foreign key constraint on parameter_id table in parameters_mapping
* In table parameters, the imos_vocabulary_id is defined as character but is set as integer in the ipostgresqlOutput component.
* the length of the parameters differ between the table definition and the ipostgresqlOutput component in the unit table. Maybe easier to define these parameters as character varying like Marty suggested and forget about these arbitrary length you need to specify later in these components.

xhoenner:
Thanks @bpasquer, changes made in https://github.com/aodn/harvesters/pull/585/commits/99aca32c3887007b716a54629f88a391445fcf7b

PR #586, opened 2018-01-10
New AODN_NSW_OEH subjob for Beachwatch temperature data

xhoenner:
@mhidas have a look at it once the corresponding pipeline is ready. Not to be merged until then.


workspace/AODN_NSW_OEH/process/AODN_NSW_OEH_harvester_0.3.item
```diff
@@ -88,7 +96,7 @@
...
...&#xA;        DROP VIEW IF EXISTS aodn_nsw_oeh_beachwatch_map CASCADE;&#xA;        DROP VIEW IF EXISTS aodn_nsw_oeh_beachwatch_data CASCADE;&#xA;&#xA;        CREATE VIEW aodn_nsw_oeh_beachwatch_map AS&#xA;          select url,&#xA;          location, &#xA;          site_name, &#xA;          min(survey_date) as survey_start_date, &#xA;          max(survey_date) as survey_end_date,  &#xA;          count(*) AS nb_surveys,&#xA;          geom&#xA;          from beachwatch_measurements bm&#xA;          join indexed_file i on i.id = bm.file_id&#xA;          group by url, location, site_name, geom&#xA;        order by location, site_name;&#xA;&#xA;        CREATE VIEW aodn_nsw_oeh_beachwatch_data AS&#xA;          select bm.id,&#xA;          url,&#xA;          file_id,&#xA;          location,&#xA;          longitude,&#xA;          latitude,&#xA;          survey_date AT TIME ZONE 'UTC' AS survey_date, &#xA;          site_name,&#xA;          sea_water_temperature,&#xA;          geom&#xA;          from beachwatch_measurements bm&#xA;          join indexed_file i on i.id = bm.file_id&#xA;        order by id;&#xA;&#xA;      &lt;/sql>&#xA;    &lt;/changeSet>&#xA;&#xA;&lt;/databaseChangeLog>&#xA;"/>
```
bpasquer: One comment on the survey_start_date and survey_end_date in the  beachwatch map and data views. In most of our data collections the time extent is stored in column names called "time_coverage_start" and "time_coverage_end". The way we store geometry and time extent has been standardized as a requirement from MARVL to allow MARVL to easily make WPS request for gridded and non gridded dataset. I think we should follow this naming 'convention' as long as it is relevant.

mhidas: @bpasquer Good point. I think for beachwatch it would make sense to have the timestamps in a column called "time" in the `beachwatch_measurements` table, and for the corresponding min/max values to be called "time_coverage_start" and "time_coverage_end".

I'm not sure we should also do this for the other _map views in this schema (single-/multi-beam). They are mostly bathymetry data, which I suppose do change with time, but on much slower timescales. So the timestamps just record the date the survey was completed, but the measurements are representative of a longer time period. Also, the single- & multi-beam products are in various formats, not NetCDF, so I don't think MARVL can do anything with them anyway.

I'll have a look at the harvester later...

bpasquer: So yes, to be more precise:
* geometry is 'geom'
* timestamps like in measurement tables are 'time' (not case sensitive)
* time extent is time_coverage_start / end

xhoenner: Thanks @bpasquer, changed all timestamp fields to comply with the vocabulary highlighted in your review --> https://github.com/aodn/harvesters/pull/586/commits/ac3cb67f7f53dfab9433b8427b4ebe57f84735d2

bpasquer:
@xhoenner you probably think I'm too picky...
* Tick "Die on error" in  tfileInputDelimited of  HarvestBeachwatchCSV job

xhoenner:
> Tick "Die on error" in tfileInputDelimited of HarvestBeachwatchCSV job

Done

PR #588, opened 2018-02-06
ANMN_ADCP_TIMESERIES: remove unique constraints in schema

mhidas:
These constraints can cause the harvest to fail and kill the whole pipeline process. We'll deal with duplicates via an independent process.


mhidas:
All the changes can be seen in [this diff](https://github.com/aodn/harvesters/pull/588/commits/b436c7fbd9f7747c7ed10a6aa7fe76709d47ae9a#diff-553959f7fbf5af4bc903b33696a06550)

PR #596, opened 2018-02-09
 ANMN_METADATA: update spatial extent in relevant metadata records

mhidas:
Set up to update spatial extent for layers:
* ABOS SOTS delayed mode
* All moorings (disabled for now as collection is not public)


mhidas:
If anyone has a chance to have a brief look at this, I'd appreciate that. Otherwise I'm happy to merge it myself and test in RC.

What I'm working towards is to replace some (eventually all?) moorings harvesters by subjobs in this "mega-harvester". That would reduce some of the complexity in the pipelines due to running multiple harvesters on the same files.

PR #604, opened 2018-03-21
ANMN_METADATA: Fix null pointer exception due to undefined context variable

mhidas:
This fixes a bug introduced in #596


mhidas:
@ggalibert FYI

Happy for anyone to post-review, if you're bored.

PR #612, opened 2018-06-20
Delete obsolete harvester ABOS_SOTS

mhidas:


mhidas:
This collection is now created by the ANMN_METADATA harvester.

PR #615, opened 2018-06-28
ANMN_METADATA: determine file real-time status internally

mhidas:
... rather than from context. This will allow the current talend jobs **anmn_delayed** and **anmn_realtime** to be merged into a single job, with a much simpler regex.


mhidas:
@xhoenner Would you mind taking a look?

mhidas:
:+1: Thanks!

PR #617, opened 2018-07-03
ANFOG_DM: implement processing of NRL Adapter missions

bpasquer:
- add creation of aodn _nrl map and data views
- modify andog_dm map and data view to exclude nrl missions
- add NRL specific iUpdateSpatialExtent component


lbesnard:
https://github.com/aodn/chef-private/blob/master/data_bags/talend/anfog_dm.json will need to be modified to add the new context

PR #624, opened 2018-08-16
Set TLS 1.2 for default on CSIRO endpoint harvesters

anguss00:


mhidas:
@anguss00 @jonescc I have re-deployed the two harvesters after this change to RC, but I still get the same error for for soop_auscpr:
```
Exception in component sWfsInput_1_GI
javax.net.ssl.SSLException: Received fatal alert: protocol_version
```
(anmn_nrs_bgc in RC seems to have an unrelated problem. It's failing to run the liquibase)

jonescc:
Looks like it needs -Dhttps.protocols=TLSv1.2 as well @anguss00

jonescc:
Make that instead of the client.protocols.   Need to read up on the difference client.protocols works for me on my version of java 7.  Suspect it changed somewhere along teh lines

anguss00:
Issue with client.protocols fixed in b05 of 1.7.0_95 - we are running b01. See https://bugs.openjdk.java.net/browse/JDK-8076369. Introducing new PR to fix using -Dhttps.protocols=TLSv1.2 (unfortunately)

PR #641, opened 2018-11-06
Fix SOOP CO2 NRT and AODN WAVE NRT mapping. Add query for variable wi…

bpasquer:
…thout defined units in unit table


bpasquer:
Found a workaround that is acceptable for AODN WAVE NRT parameter mapping see https://github.com/aodn/content/issues/377 . 
However this is a temporarily fix  for SOOP CO2 NRT mapping  until H2O_Raw_PPT unit is added to the vocab

PR #645, opened 2018-11-14
New harvester PLANKTON_PRODUCTS

mhidas:
This will harvest all data products produced by the plankton group at CSIRO (for now only CPR phytoplankton products -- others to be added later)


mhidas:
If anyone has time & appetite to do a quick review, I would be happy for some feedback. Independently of that, I would like to get this into RC next week.

@xhoenner @bpasquer @lbesnard

bpasquer:
Thanks @xhoenner, I don't have much time to look at it this week.

mhidas:
Thanks @xhoenner

xhoenner:
Looking good @mhidas, my only question: why is the 'Delete metadata for deleted resources' subjob deactivated?

mhidas:
Thanks @xhoenner. I don't think deletions will be needed for this harvester. Each product will be just one csv file, updated every now and then with a new one. Currently I don't see any need for the deletion function. There's more deveolopment to be done on this harvester later, so I'll rethink that at the time, but for now this should do.

PR #650, opened 2018-12-06
AODN_WAVE_DM :  generate a single layer for the collection instead of…

bpasquer:
… one

per institution
- delete the institution specific metadata
- create  new _map and _data view


bpasquer:
@lbesnard  Can you please take care of this one please

PR #659, opened 2019-01-14
AODN WAVE DM Reference to variable TIME in the _data view ORDER BY us…

bpasquer:
…ing the timezone function


bpasquer:
@lwgordonimos Is the change in the ORDER BY  expression what you recommended to do to fix the issue with the AODN WAVE DM _data layer?

lwgordonimos:
In any case, the index needs to match the sort fields exactly.

One option is what you've done, and made the sort fields match the current index, which uses the function output.

The other option would be to leave the query and change the **index** so that it indexes on the raw column (and not the result of the timezone function), e.g.

```sql
CREATE INDEX measurements_t_idx
  ON aodn_wave_dm.measurements
  USING btree
  (file_id, "TIME");
```

I would lean towards the latter because the index is more generically useful for anything involving the TIME column and it's always simpler to not deal with functional indexes if they aren't strictly necessary.

I think after a bit of testing, either will work.

PR #670, opened 2019-02-06
 ANFOG-DM: replace reference to NRL by UWA

bpasquer:


lbesnard:
we also need to change the data bags right ?

bpasquer:
indeed ! 
https://github.com/aodn/chef-private/pull/3147

PR #680, opened 2019-02-22
SOOP TMV - REALTIME

lbesnard:
* Use SOOP TMV harvester (previously DM only) to process RT files
 * realtime files must have realtime string in their path
 * add realtime boolean column on visualisation wms
 * use realtime boolean to separate DM and NRT views


bpasquer:
it's neat 👍

PR #685, opened 2019-03-04
PARAMTERS MAPPING improvement

bpasquer:
- add column comment to parameter table
 - select column comment in relevent views


bpasquer:
Fix https://github.com/aodn/content/issues/404

PR #686, opened 2019-03-05
SRS OC DALEC - add to existing harvester new collection of dalec data…

lbesnard:
… mounted on various soop; renaming of views ...


lbesnard:
data bags on chef-private need to be modified to add new uuid

lbesnard:
https://github.com/aodn/chef-private/pull/3160

PR #694, opened 2019-03-13
SOOP_AUSCPR: fix checking for empty Metadata_UUID_biomass (null or "")

mhidas:


mhidas:
This was  failing in  RC because that context variable is currently  NULL,  and it  was  trying to use  a string method to test  whether it was  equal to "". Now  it  checks  for a null  value first.

PR #696, opened 2019-03-15
Rename old tPostgresqlOutput_1 to prevent naming clash with renamed i…

jonescc:
…PostgresqlOutput_1


jonescc:
needs another update

PR #704, opened 2019-03-28
PLANKTON_PRODUCTS: fix up harvesting of raw tables

mhidas:


mhidas:
This should fix https://github.com/aodn/issues/issues/395 (as suggested [here](https://github.com/aodn/issues/issues/395#issuecomment-476933990) - thanks @jonescc )

PR #711, opened 2019-04-10
SOOP TMV - add pkey to visualisation_wms

lbesnard:


lbesnard:
simple liquibase change

PR #714, opened 2019-05-03
Fixes incorrect timestamp calculation reported in aodn/content/416

jonescc:
for anmn_ts_timeseries harvester

Problem occurred when using Calendar instance to round to nearest second.  There was an associated time zone of AEDT which got converted to AEST when adding 500ms for some reason.

Remove dependency on netcdf library for performing this calculation.

Add unit testing code to routine.

### DIFF
--- a/workspace/AATAMS_ACOUSTIC_REPORTING/code/routines/NetCDFUtils_0.1.item
+++ b/workspace/AATAMS_ACOUSTIC_REPORTING/code/routines/NetCDFUtils_0.1.item
@@ -1,83 +1,226 @@
 package routines;
 
+import java.text.DateFormat;
+import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.TimeZone;
 
-import ucar.nc2.units.DateUnit;
+import routines.system.FastDateParser;
 
 /*
  * user specification: the function's comment should contain keys as follows: 1. write about the function's comment.but
  * it must be before the "{talendTypes}" key.
- * 
+ *
  * 2. {talendTypes} 's value must be talend Type, it is required . its value should be one of: String, char | Character,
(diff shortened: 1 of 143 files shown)


jonescc:
Note: this routine is currently included in each harvester that uses it.  The fix here just updates that code, but its also possible to rewrite this routine as a plugin system routine so the code would only need to be fixed in one place.  It would then be copied to the harvester whenever it was opened/built.  Need a backlog item for that as there is some work involved.

lbesnard:
@jonescc how do we test/deploy this via jenkins since it affects a lot of harvesters and will blow away the limit of 25 or so of harvester pipelines to be displayed on jenkins?

jonescc:
I'll check, but I can't see why there needs to be a limit anymore - all the artifacts are being stored on s3 now.

jonescc:
The alternative is separate pull requests for each harvester.

jonescc:
@lwgordonimos, any reason why we can't increase the number of builds kept for talend jobs now?   Its currently set to 40.   I've set the talend 7 build jobs to retain up to 200 without any issues.

lwgordonimos:
@jonescc no reason any more since the build artifacts are now 100% S3. Storing a single build is now <200kB on the Jenkins server, so should be fine!

jonescc:
@lbesnard - jenkins has been modified to display up to 200 previous builds.

lbesnard:
thanks. ill wait for all the po's to be back next week before merging this so we can push to prod the previous versions first

lbesnard:
@jonescc can we rebase this PR against master ? or shall we wait until all TOS 7 harvesters are on prod ?

jonescc:
Will need to rebase against the tos 7 branch which will eventually be merged into master.  No point doing it in master as the code it changes is going to change.

jonescc:
Rebased against master now that tos-7-1-1 has been merged - should be OK to go when ready.

Issue #734, opened 2019-05-22
auv harvester not clearing deleted data out of auv_trajectory_map


jonescc:
See https://github.com/aodn/utilities/pull/127

lbesnard:
this was fixed in above PRs

PR #735, opened 2019-05-22
AUV - fix hrvester to remove deleted data

lbesnard:


lbesnard:
https://github.com/aodn/harvesters/issues/734

@jonescc this should fix it. There was a fair bit of merging of files belonging to a same dive being done, in order to avoid duplication of data

jonescc:
Couldn't build this.  Looks like there are some issues with the delete_clean_data tRunJob:

image

lbesnard:
@jonescc I did a small change (fileid was set as string rather than long). a bit hard to test, but I could at least run a talend liqui on it on my pobox

jonescc:
Getting a null pointer exception running this on 6-nec-hob when deleting a file.  Looks like we're referring to ((Long)globalMap.get("iDeletedFileList_1_FILE_ID")).intValue() in delete_clean_data.   The global map isn't shared between jobs so this is undefined in this job.  Should be refering to fileid which is being passed in.

Also, not sure if it was like this before or not, but, you wouldn't normally put another operation which could fail in the tPostJob component before the tPostgresqlClose.  The purpose of the tPostJob/tPostgresqlClose combination is to ensure the database connection is ALWAYS closed.   Putting a tPostgresqlRow before the close which can fail like it did here means that this won't necessarily be true anymore.

jonescc:
Also, any reason why the mergeMeasurements and createWMS components are included in the tPostJob (which to me is more about ensuring resources are properly cleaned up) rather than in the main flow for this job?

PR #736, opened 2019-05-22
SOOP TMV NRT - remove order by on WFS layer (slowing download) + add …

lbesnard:
…missing columns to WFS (product_type + vessel_status)


lbesnard:
@bpasquer same as the previous one. But I forgot to add those columns to the WFS making the filters not working as expected

PR #746, opened 2019-05-27
AUV - improve harvester

lbesnard:
* delete data was not working
 * mod tables "table_master" and "merged_measurements" to have "timestamp with timezone"

PR #751, opened 2019-05-30
AUV - fix time with timezone on wms table

lbesnard:
@jonescc fixing issue https://github.com/aodn/harvesters/pull/746#issuecomment-496724376


jonescc:
Looks good.  Will test on 6-nec-hob

PR #766, opened 2019-07-24
PLANKTON_PRODUCTS: migration to Talend 7

mhidas:


mhidas:
@jonescc Here you go

jonescc:
Doesn't look to me like the pom directory is actually required - its generated whenever I open the job.

mhidas:
Ok, I've removed poms and added these directories to .gitignore (along with temp)

mhidas:
Thanks @jonescc . FYI this has been deployed to RC and runs just fine.

PR #767, opened 2019-07-25
Argo - modify history_parameter char size to 64 from profile_history …

lbesnard:
…table

performed with TOS7


lbesnard:
@jonescc a simple modification but since it's done with TOS7, I want to be sure we 've got the workflow right

jonescc:
Looks OK to me - requires https://github.com/aodn/chef/pull/847 to be merged and deployed before we can build this in jenkins as we don't currently have the new postgres driver installed there.

PR #773, opened 2019-07-31
PLANKTON_PRODUCTS: update geometry in nrs_ind table

mhidas:


mhidas:
Just adding a component I left out in the original PR.

PR #774, opened 2019-07-31
Prototype for building harvesters on the new jenkins

akashisama:
Includes a Dockerfile for running builds and a Jenkinsfile for the AUV harvester (we'll need to make more Jenkinsfiles later, one for each harvester)

Can be tested in conjunction with https://github.com/aodn/cloud-deploy/pull/474


Dockerfile
```diff
@@ -0,0 +1,99 @@
...
+
+# Download and install talend maven repo
+RUN cd /tmp \
+  && curl -SL  https://s3-ap-southeast-2.amazonaws.com/imos-binary/static/talend/talend-maven-repo.zip -O \
+  && mkdir -p /opt/TOS_DI-20181026_1147-V7.1.1/configuration/.m2 \
+  && unzip -d /opt/TOS_DI-20181026_1147-V7.1.1/configuration/.m2 /tmp/talend-maven-repo.zip \
+  && rm /tmp/talend-maven-repo.zip
+
+# Install xulrunner
+RUN cd /tmp \
```
lwgordonimos: I'm assuming this is needed because of the GUI hacks we are doing to build it?

Dockerfile
```diff
@@ -0,0 +1,99 @@
...
+  && curl -SL https://s3-ap-southeast-2.amazonaws.com/imos-binary/jobs/talend7_components_edge/2/components-1.0.0-SNAPSHOT.zip -O \
+  && unzip /tmp/components-1.0.0-SNAPSHOT.zip -d /opt/TOS_DI-20181026_1147-V7.1.1/talend-components \
+  && rm -rf /tmp/components-1.0.0-SNAPSHOT.zip /tmp/target
+
+# Add JAVA_HOME and add to path as required by TOS
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
+ENV PATH $JAVA_HOME/bin:$PATH
+
+RUN mkdir /tmp/.talend7-workspace
```
lwgordonimos: This should all be running as the builder user (this is going to cause builds to run as root), and shouldn't need 777 anywhere. If the builder user needs write access somewhere, it can be explicitly granted to just that user.

workspace/AUV/Jenkinsfile
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+        stage('container') {
+            agent {
+                dockerfile {
+                    additionalBuildArgs '--build-arg BUILDER_UID=${JENKINS_UID:-9999}'
+                }
+            }
+            environment {
+                TALEND_DIR = "/opt/TOS_DI-20181026_1147-V7.1.1"
```
lwgordonimos: If these are static, they should be in the Dockerfile as static environment variables, since they aren't related to executing the build, they are attributes of the Docker *image*.

workspace/AUV/Jenkinsfile
```diff
@@ -0,0 +1,68 @@
+#!groovy​
+
+pipeline {
+    agent none
+    stages {
+        stage('validation') {
+            agent { label 'master' }
+            steps {
+                //Moving in to the directory to execute the commands
+                dir('workspace/AUV') {
```
lwgordonimos: I feel like the triggering logic should still be separated out from from the build definition rather than jammed in here.

The current Jenkins separates this with the separate trigger job to do the nasty comparison stuff, which I think is still correct. The trigger job can determine which subdirectories have changed, and fire off the per-harvester builds accordingly.

The declarative Jenkinsfiles will be much better if they can focus on "how to build" not "why to build".

Dockerfile
```diff
@@ -0,0 +1,99 @@
...
+# Dockerfile to build image for running Talend Open Studio 7.1.1
+#################################################################
+
+FROM ubuntu:18.04
+MAINTAINER AODN
+
+ARG BUILDER_UID=9999
+
+# Install Java 8.
+RUN \
```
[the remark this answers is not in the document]
akashisama: Yeah sounds reasonable, these installations are copied from the codegen talend. I didn't try to make it more efficient, since it's just a prototype right now.

Dockerfile
akashisama: Not sure, brought over from the codegen dockerfile. I tried to do a build removing all the package installations that didn't seem necessary but it failed and I figured it wasn't worth the effort (during a prototype stage) to remove them one by one to find out which ones were and weren't necessary.

Dockerfile
akashisama: Mmm yeah I originally tried creating a builder user à la our other dockerfiles and setting the permissions for just that user... but Jenkins doesn't actually seem to run from the builder user? If you look at the build log for any of our jobs it shows the docker command begins ```docker run -t -d -u 498:497``` but 498:497 isn't the id of the builder user/group...

workspace/AUV/Jenkinsfile
akashisama: I'm not really using these as environment variables, I'm just setting them here to access below in the jenkinsfile when I call the talend command. This is probably bad, I expect there's a more correct way to set variables for later use the jenkinsfile (which I want to do to make it more readable).

workspace/AUV/Jenkinsfile
akashisama: Yeah it would be good to keep this sort of logic out of the jenkinsfile, but I couldn't see how to do that without moving away from the "multibranch pipeline" approach we're using for other jobs. We should probably explore this further in a future iteration now we've got a working prototype.

Dockerfile
lwgordonimos: 498 is what the jenkins user usually gets, which is the purpose of the "docker build  arguments" in the Jenkinsfile, to ensure the UIDs line up. Anyway, I think it's easy fixed once it's out of prototype mode.

lwgordonimos:
I think we can consider the prototype complete, but I don't think this should be merged to master in this branch. Be better if this was a draft PR, and we can then work on it in anger before merging into the code repo(s) (as opposed to the config management repo).

Dockerfile
akashisama: :( JENKINS_UID and the other environment variables intended to be set at https://github.com/aodn/cloud-deploy/blob/301bf2af5aed127cbef43d645c2e5b9808fe4b6a/ansible/roles/jenkins_config/files/base-configuration-as-code.yml#L6 aren't accessible from the jobs. 

Doesn't seem like EnvInject plugin works with pipeline:

> ### Known incompatibilities
> #### Jenkins Pipeline
> Even though it is possible to set up the EnvInject Job Property and build step in Pipeline, the plugin does not provide full compatibility with Pipeline Plugin.
> 
> Supported use-cases:
> 
> Injection of EnvVars defined in the "Properties Content" field of the Job Property
> These EnvVars are being injected to the script environment and will be inaccessible via the "env" Pipeline global variable
> Please note there is also a known compatibility issue with Durable Task Plugin 1.13
> All other use-cases of the plugin may work incorrectly in Jenkins Pipeline. Please see JENKINS-42614 for more information about unsupported use-cases. There is no short-term plan to fix the compatibility issues though any pull requests to the plugin will be appreciated.

akashisama:
Replaced by https://github.com/aodn/harvesters/pull/777 and https://github.com/aodn/harvesters/pull/776

PR #776, opened 2019-08-27
New jenkinsfiles for each harvester

pmbohm:
##### Note:
Files created via script, so there wont be typo's to look for.


workspace/AATAMS_ACOUSTIC_QC/Jenkinsfile
```diff
@@ -0,0 +1,67 @@
...
+    environment {
+        JOB_WORKSPACE = 'workspace/AATAMS_ACOUSTIC_QC'
+        JOB_NAME = 'AATAMS_ACOUSTIC_QC'
+    }
+    stages {
+        stage('validation') {
+            agent { label 'master' }
+            steps {
+                //Moving in to the directory to execute the commands
+                dir(${JOB_WORKSPACE}) {
```
akashisama: The ${} syntax is only for within strings, it won't work here.
```suggestion
                dir(JOB_WORKSPACE) {
```

lwgordonimos: Also, pretty sure it's exposed in the groovy scope as `env.JOB_WORKSPACE`

workspace/ANFOG_DM/Jenkinsfile
```diff
@@ -4,7 +4,7 @@ pipeline {
     agent none
     environment {
         JOB_WORKSPACE = 'workspace/ANFOG_DM'
-        JOB_NAME = 'ANFOG_DM'
+        JOB_NAME = 'mainANFOG_dm'
```
akashisama: Think you must have caught the wrong files for these ANFOG ones: top level properties file is not an exception https://github.com/aodn/harvesters/blob/master/workspace/ANFOG_RT/process/ANFOG_RT/ANFOG_RT_harvester_0.1.properties

PR #777, opened 2019-08-29
Add docker file for building harvesters

akashisama:
Based on prototype at https://github.com/aodn/harvesters/pull/774

Part of https://github.com/aodn/backlog/issues/1233


lwgordonimos:
Just had a go at building a locally from the Dockerfile. Looks good, I just made a couple of tweaks.

- wget vs curl for consistency with other Dockerfiles
- got rid of a couple of unnecessary operations, `chown` and copying the single JAR file after downloading
- added .dockerignore - we need to do this for everything, because it is a waste zipping the whole repos when the Dockerfile doesn't depend on anything *in* the repo
- added docker-compose.yml to make it easier for local dev work

akashisama:
@lwgordonimos Yep I like those changes, happy for you to merge.

PR #783, opened 2019-09-24
Create top level jenkinsfile...

akashisama:
...which works out which harvesters have changed and triggers them to build

https://github.com/aodn/backlog/issues/1269

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,7 +14,7 @@ pipeline {
                     
                     projects = sh(returnStdout: true, script: "git diff --name-only ${base } ${GIT_COMMIT} | awk -F '/' '{print \$2}' | uniq").trim().split()
                     for(project in projects) {
-                        echo project
+                        echo "Found changes in project ${project}"
                     }
                     
                 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,53 @@
+pipeline {
+    agent none
+    stages {
+        stage ('discover_changed_harvesters') {
+            agent any
+            steps {
+                script {
+                    if (env.GIT_PREVIOUS_COMMIT  == null) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,53 @@
...
+            }
+        }
+        stage('build') {
+            agent any
+            steps {
+                script {
+                    for(project in projects) {
+                        projects_directory = "workspace"
+                        harvester_job_name=sh(returnStdout: true, script: 
+                        "find $projects_directory/$project/process -regex '.*_harvester_[0-9]+\\.[0-9]+\\.item' | sed -e 's/.*\\///g' -e 's/_[0-9]\\+\\.[0-9]\\+\\.item//g' | head -n 1").trim()
```
akashisama: Would be nice to groovyfy this bash but the groovy I came up with requires script approval :(
So I'll just leave it bashy for now.
``` org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use new groovy.util.FileNameByRegexFinder ```

``` Groovy
def finder = new FileNameByRegexFinder()
processFile = finder.getFileNames('/tmp/harvesters/workspace/AUV/process', '.*_harvester_[0-9]+\\.[0-9]+\\.item')[0]
	
echo(processFile)
if (!processFile) {
	echo "Skipping $project. No job with '_harvester' suffix found"
	continue
}

harvester_job_name = processFile.split('/')[-1].replaceAll('_[0-9]+\\.[0-9]+\\.item$','')
```

akashisama:
Opening this up, ready for review. Leigh looked at it before he left and thought it seemed good, but he didn't have a chance to run it up.

Jenkinsfile
lwgordonimos: @akashisama the utility code type stuff should ideally be in the global Groovy library anyway rather than in Jenkinsfile. That may avoid the script approval requirement?

PR #785, opened 2019-09-25
Revert "Update Python using python-modernize -n -w -p on this repo"

lbesnard:
Reverts aodn/harvesters#780


lbesnard:
@pmbohm actually, I don't think this PR is needed. I'm 100% sure that this python file is not used at all by this harvester. Opening the harvester via talend, I can't see any reference to it. And this code actually exists in data-services. So maybe the harvester part was not cleaned up properly

pmbohm:
@lbesnard Should the file be removed then rather than reverted?

lbesnard:
Yeap, I was thinking of doing this in a separate PR, but you're right

PR #788, opened 2019-10-16
update spatial extent for anmn_am_nrt

evacougnon:
data_bags needs to be updated before this can be merged

Also, waiting on [this](https://github.com/aodn/issues/issues/555) issue to be fixed so we can fully test it in the pipeline box


mhidas:
@evacougnon @ocehugo 
A solution to aodn/issues/issues/555 has been identified (see https://github.com/aodn/chef/pull/899), so you should now be able test this in the pipeline box.

PR #789, opened 2019-10-21
SRS SURFACE WAVES - mod harvester to process both altimetry and scatt…

lbesnard:
…erometry files


lbesnard:
@atkinsn when merging this PR, we need to modify the metadata record link to the map view on RC first
from 
srs_surface_waves_map to srs_surface_waves_altimetry_map

PR #793, opened 2019-10-29
remove aatams_acoustic_qc harvester as pipeline is decommissioned

evacougnon:
Part of the aatams_acoustic_qc pipeline decommission [po-backlog](https://github.com/aodn/PO-Backlog/issues/370)

Other PR related to this issue to decommission the pipeline:

- on [chef-private](https://github.com/aodn/chef-private/pull/3375)
- and [data-services](https://github.com/aodn/data-services/pull/957/files)


lbesnard:
I don't think we should get rid of this harvester. We probably will be able to use it in the future without having to change anything

ggalibert:
We can remove it (it's not lost anyway) and in the future we won't be following this method if we want to do it again.

PR #806, opened 2019-12-02
Update python3 shebang

anguss00:


workspace/PLANKTON_PRODUCTS/code/python/preproc.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
```
mhidas: This script was only used locally on my machine when I was developing this harvester. It is not currently used in any RC or prod environment. However it could be handy if a pipeline is ever developed for this collection, so worth saving it here and keeping it updated.

PR #815, opened 2020-04-07
Sattag dm harvester eva

lbesnard:


lbesnard:
@evacougnon this is ready to be reviewed. Note that I still need to had the metadata update component at the end of the harvester

* All tables are filled
* added qc_ssm_outputs view
* thorough check of the table content

I don't quite understand the CTD csv file. As an example, there are multiple entries of the same time value: "2019-03-02T15:20:00Z" within the CSV file.


--------------
ISSUE with tConvertType component
--------------
The tConvertType component is behaving in a way I don't understand. As an example, let's look at the Harvest_ctd subjob which is harvesting the following subsetted file: 
[ctd_wd11.csv.zip](https://github.com/aodn/harvesters/files/4442231/ctd_wd11.csv.zip)
The job is reading for example the temp.dbar column (values are "4,10,20,30,50,75,98,100,110,120,140,150,200,236,282,294") which needs to be normalised. The normalisation seems to work fine, however after going through tConvertType component, the first value __4__ of the normalised "vector" is missing... why?
If I un-tick the ```Auto Cast``` option from the component and set all the variables manually I get the correct output number. However, this doesn't solve the issue. Why is this happening?

I have a similar issue in the Haulout, dive and other subjobs, where the number of rows is different in input and output with the tConvertType_1 component. See
talend_mystery

The following zip file contains a group of files to be added to the fileList of files to harvest
[wd11.zip](https://github.com/aodn/harvesters/files/4442240/wd11.zip)
@jonescc FYI

evacougnon:
> All tables are filled

Looks okay to me, except for the missing value due to the issue with `tConvertType`, I hope @jonescc can help us with that. 

I've just noted one minor thing, can we get the same precision on the `qc_ssm_x`, `qc_ssm_y`, `qc_ssm_x_se`, `qc_ssm_y_se`, that what is available in the csv files? 


> added qc_ssm_outputs view

when running the harvester, I get: 
```sql
SEVERE 4/16/20 11:15 AM:liquibase: Change Set changelog.xml::Recreate_aatams_sattag_qc_dm_ssm_data_view::ecougnon failed.  Error: Error executing SQL CREATE VIEW aatams_sattag_qc_dm_ssm_data AS
		 SELECT 
		    md.sattag_program,
		    a.device_id,
		    md.device_wmo_ref,
		    md.ptt,
		    md.tag_type,
		    md.common_name,
		    md.release_site,
		    md.state_country,
		    md.age_class,
		    md.sex,
		    timezone('UTC'::text, a."timestamp") AS "timestamp",
		    a.*
		   FROM aatams_sattag_qc_dm.qc_ssm_outputs a
		     JOIN aatams_sattag_qc_dm.metadata md ON a.device_id = md.device_id
		  ORDER BY a.device_id, (timezone('UTC'::text, a."timestamp")): ERROR: column "device_id" specified more than once
```
and a bit further `ERROR: column "timestamp" specified more than once`
Shouldn't we need to list all the columns? To make it work I had to replace `a.*` by:
```sql
		    a.qc_ssm_lon,
		    a.qc_ssm_lat,
		    a.qc_ssm_x,
		    a.qc_ssm_y,
		    a.qc_ssm_x_se,
		    a.qc_ssm_y_se
```

> I don't quite understand the CTD csv file. As an example, there are multiple entries of the same time value: "2019-03-02T15:20:00Z" within the CSV file.

@bpasquer do you know something about that (see below)?

I'm not entirely sure either! I can't find anything specific in the documentation I have!...
What I can understand from the csv files is that they are different sensors on the same tag:
- temp + salt
- fluoro (only measured until 200 dbar)
- photo (as far as I know we do not harvest this field)

Few problems with that:
- where you have temp+salt values, fluoro is empty and vise versa. Therefore, I wonder if they should be part of the same profile? 
- the pressure associated to `fluoro.vals` is recorded in `fluoro.dbar`, so `fluoro.dbar` needs to also be harvested as it seems the fluoro is not measured at the same pressure intervals than the other variables (currently we harvest `temp.dbar` as the pressure for all)
- in the csv, `n.temp`, `n.salt`, `n.fluoro` indicate how many pressure/temp (pressure/salt, pressure/fluoro) pairs we have for one row, that might be useful if we want to filter

I see 2 approaches:
- we harvest as it is now, i.e. 1 profile = 1 row in the csv file, which means we'll have for the same timestamp/location several profiles (1- with only `temp` and `salt`; 2- with only `fluoro`; 3- all empty as we do not harvest `photo`
- we combine the profiles

jonescc:
> The normalisation seems to work fine, however after
going through tConvertType component, the first value **4** of the
normalised "vector" is missing... why?

The problem I get (after fixing the device_id issue) looks to be an
attempt to convert "NA" in some of the input csv columns into a numeric
value.

Without Die On Error set in TConvertType, any conversion errors will
result in the record being skipped.  The conversion error message is printed to stderr - 'For input string: "NA"' but this probably wasn't very informative.  If you set Die On Error in the component properties, any conversion errors will throw an exception, the file not will not be processed and an exception stack trace will be displayed making it much clearer what the problem is.

```
Exception in component tConvertType_1 (Harvest_ctd)
java.lang.NumberFormatException: For input string: "NA"
    at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2043)
    at sun.misc.FloatingDecimal.parseFloat(FloatingDecimal.java:122)
    at java.lang.Float.parseFloat(Float.java:451)
    at java.lang.Float.valueOf(Float.java:416)
    at routines.system.TypeConvert.String2Float(TypeConvert.java:4476)
    at aatams_sattag_qc_dm.harvest_ctd_0_1.Harvest_ctd.tFileInputDelimited_1Process(Harvest_ctd.java:7639)
    at aatams_sattag_qc_dm.harvest_ctd_0_1.Harvest_ctd.tDBRow_1Process(Harvest_ctd.java:1240)
    at aatams_sattag_qc_dm.harvest_ctd_0_1.Harvest_ctd.tDBConnection_1Process(Harvest_ctd.java:1005)
    at aatams_sattag_qc_dm.harvest_ctd_0_1.Harvest_ctd.iIncludeSdiLibraries_1Process(Harvest_ctd.java:764)
    at aatams_sattag_qc_dm.harvest_ctd_0_1.Harvest_ctd.runJobInTOS(Harvest_ctd.java:11349)
    at aatams_sattag_qc_dm.harvest_ctd_0_1.Harvest_ctd.runJob(Harvest_ctd.java:11133)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.iNewOrModifiedFileList_2Process(AATAMS_SATTAG_QC_DM_harvester.java:4321)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.iNewOrModifiedFileList_1Process(AATAMS_SATTAG_QC_DM_harvester.java:3887)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.iDeletedFileList_1Process(AATAMS_SATTAG_QC_DM_harvester.java:3278)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.iPostgresqlDbUpdate_1Process(AATAMS_SATTAG_QC_DM_harvester.java:2713)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.iUpdateIndex_1Process(AATAMS_SATTAG_QC_DM_harvester.java:1874)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.runJobInTOS(AATAMS_SATTAG_QC_DM_harvester.java:9391)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.main(AATAMS_SATTAG_QC_DM_harvester.java:9029)
Exception in component tRunJob_2 (AATAMS_SATTAG_QC_DM_harvester)
java.lang.RuntimeException: Child job running failed.
java.lang.NumberFormatException: For input string: "NA"
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.iNewOrModifiedFileList_2Process(AATAMS_SATTAG_QC_DM_harvester.java:4349)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.iNewOrModifiedFileList_1Process(AATAMS_SATTAG_QC_DM_harvester.java:3887)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.iDeletedFileList_1Process(AATAMS_SATTAG_QC_DM_harvester.java:3278)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.iPostgresqlDbUpdate_1Process(AATAMS_SATTAG_QC_DM_harvester.java:2713)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.iUpdateIndex_1Process(AATAMS_SATTAG_QC_DM_harvester.java:1874)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.runJobInTOS(AATAMS_SATTAG_QC_DM_harvester.java:9391)
    at aatams_sattag_qc_dm.aatams_sattag_qc_dm_harvester_1_0.AATAMS_SATTAG_QC_DM_harvester.main(AATAMS_SATTAG_QC_DM_harvester.java:9029)
[statistics] disconnected
``

If you don't want conversion errors almost silently ignored and the record skipped, I'd turn Die On Error on.

You'll need to do something to handle the "NA" values - e.g. use a tMap to convert it to null for example.

Looking at the input csv files the problem values should be pretty obvious.

lbesnard:
> I've just noted one minor thing, can we get the same precision on the `qc_ssm_x`, `qc_ssm_y`, `qc_ssm_x_se`, `qc_ssm_y_se`, that what is available in the csv files?

I'm not sure I get this. The columns (for these variables) are alway double across all the various tables. I checked and it looks like the precision is the same in the CSV and the database.

lbesnard:
@jonescc thanks a lot, it works well!

lbesnard:
@evacougnon this version fixes all the various issues (hopefully)

lbesnard:
another version @evacougnon  with what we chatted about

lbesnard:
@evacougnon the harvester is ready to be merged

lbesnard:
@evacougnon following our discussion on thurday on fri, can you remind me what do you want me to change? adding a metadata column somehwere

evacougnon:
`dive_start` and `dive_end` have appeared on the new metadata.csv, between the columns `actual_mass` and `qc_start_date` . We need to add them to the schema (`metadata` table).

lbesnard:
@evacougnon ready for you to have a look

PR #817, opened 2020-04-22
ATF reporting - add db port context; remove new postgres driver versi…

lbesnard:
…on in connection


lbesnard:
This is the line which was probably failing on RC. The harvester is untestable on the stack as it would consume my machine for a day (FYI @ccmoloney , but this is really not an issue)

Thanks @lwgordonimos  for find this
https://github.com/aodn/issues/issues/706#issuecomment-616978740

PR #820, opened 2020-05-14
ATF reporting - improve species view performance - re-initialised ins…

lbesnard:
…tallation summary view to prior version


lbesnard:
This PR does the following
 * performance improvement to on "aatams_acoustic_species_all_deployments_view" to use latest query from @jonescc 
" installation_summary" view is pushed back to Xav's version

PR #824, opened 2020-05-27
SAR - add is_realtime boolean to improve performances + other idx ava…

lbesnard:
…ilable as filters on portal


lbesnard:
list of changes
https://github.com/aodn/content/issues/456#issuecomment-634382893

lbesnard:
ready to be merged

PR #826, opened 2020-06-10
AATAMS SATTAG QC - remove dive_start and dive_end from metadata subjo…

lbesnard:
…b. add actual_mass column to metadata subjob


lbesnard:
some time values were set as NA in this new set of csv files. It's a pretty fun task to debug talend and handle such a simple thing as parsing date as null values... I'm going to pat myself as I find incredible I managed to make it work... by myself. Long live Talend

the trick was to do this:
```
row2.release_date.equals("NA")? null: TalendDate.parseDate("yyyy-MM-dd'T'HH:mm:ssZ",StringHandling.EREPLACE(row2.release_date,"Z","+0000")) 
```
since what was failing was the string handling  ereplace bit in the case the time value was set to null or NA

lbesnard:
@evacougnon This is now fixed

PR #830, opened 2020-06-26
SOOP TRV - add trip_id index on master_merged, cleaning of duplicate …

lbesnard:
…data in various measurements_* tables; change query merging data together


lbesnard:
will have to test these changes on RC against a replica of the prod DB as this cannot be done on the stack

PR #836, opened 2020-07-23
SRS SAR - re-add missing size column in WMS layer

lbesnard:


lbesnard:
very simple liquibase change to add ",size" in the map view for FV00 and FV01 layer currently broken on prod

run fine on stack with a prod schema restore

PR #837, opened 2020-07-28
create views for aatams_sattag_qc_dm/nrt

evacougnon:
https://github.com/aodn/data-services/pull/1007 will need to be merged before


bpasquer:
Casting variables to text , ie` 'cf_standard_name'::text`  overcomplicates the query. You can get rid of most of them, except of course when you want to add specific characters like `#`

ana-berger:
Closing this PR as it was substituted by [PR#903](https://github.com/aodn/harvesters/pull/903) to solve the conflicts.

ana-berger:
> `SELECT ((((((((('# data_column_name'::text || ','::text) || 'cf_standard_name'::text) || ','::text) || 'imos_vocabulary_name'::text) || ','::text) || 'unit_name'::text) || ','::text) || 'unit_short_name'::text) || ','::text) || 'uv.definition'::text`
> a minor comment ( valid for both queries)
> on the in the final Select query : it is actually setting the column names for the different element of the parameters definition.
> so just to make it cleaner, remove the `uv.` in front of `definition`, cause the user won't understand what is refers to

This suggestion was addressed in [PR#904](https://github.com/aodn/harvesters/pull/904).

PR #838, opened 2020-07-29
plankton_products_metadata_summary adding new view

ana-berger:


ana-berger:
example of the metadata headers:
[nrs_ind_map.txt](https://github.com/aodn/harvesters/files/4993614/nrs_ind_map.txt)
Also, I found some issues with commas in the parameters_mapping.csv file and changed it. It is a new commit in #https://github.com/aodn/data-services/tree/ana-berger-patch-1

ana-berger:
> * I think finding the info about a particular column would be much easier if they were ordered by column name first, then product. So in the query, you could have `ORDER BY variable_name, product` (those two together should be unique, so no need to list all the other columns).

I agree, just changed that. I was not happy with having it ordered by the product name indeed, it was looking confusing. Better now!

> * Many columns are missing, e.g. SampleDateUTC, SampleDateLocal, and all the ShannonXDiversity ones... I can see them in the parameters_mapping table, but they seem to be excluded from the view, so we'll have to have a closer look at that. I'm guessing it's because all these only have a long_name and everything else is NULL in the parameters table.

Ok, I see now. There are 19 variables missing. All of them actually have no "unit_id" in the parameters_mapping table. That's what they have in common. Some of them actually have "cf_standard_name", so it is not about having everything else null. One example is the SampleDateUTC; now they have 'time' as "cf_standard_name" but they still do not appear in the view.

ana-berger:
> > * Many columns are missing, e.g. SampleDateUTC, SampleDateLocal, and all the ShannonXDiversity ones... I can see them in the parameters_mapping table, but they seem to be excluded from the view, so we'll have to have a closer look at that. I'm guessing it's because all these only have a long_name and everything else is NULL in the parameters table.
> 
> Ok, I see now. There are 19 variables missing. All of them actually have no "unit_id" in the parameters_mapping table. That's what they have in common. Some of them actually have "cf_standard_name", so it is not about having everything else null. One example is the SampleDateUTC; now they have 'time' as "cf_standard_name" but they still do not appear in the view.

The variables that have no unit_id in the parameters_mapping table are being ignored to generate the view. They are integer instead of text. The query is ignoring the variables without a value for the integer 'unit_id'. I imagine is something here 
(in bold):

WITH p AS (
      SELECT '# ' || product || ',' || variable_name || ',' || cf_standard_name || ',' || long_name || ',' || imos_vocabulary_name || ',' || uv.name || ',' || uv.short_name || ',' || definition
      FROM parameters_mapping pm
      **LEFT JOIN units uv ON pm.unit_id = uv.id**
      LEFT JOIN parameters p ON p.unique_id = pm.parameter_id
      WHERE facility = 'SOOP/ANMN' AND subfacility = 'CPR/NRS'
      ORDER BY variable_name , product)
      SELECT '# data_product_name' || ',' || 'data_column_name' || ',' || 'cf_standard_name' || ',' || 'long_name' || ',' || 'imos_vocabulary_name' || ',' || 'unit_name' || ',' || 'unit_short_name' || ',' || 'definition'
    UNION ALL
      SELECT * FROM p WHERE "?column?" IS NOT NULL;

mhidas:
From looking at that query, the problem seems to be the condition `WHERE "?column?" IS NOT NULL`. I'm not sure why this condition is there in the other metadata views, but at least here I don't think you need it. 

The `LEFT JOIN` you highlighted in bold will result in a table with every column of `parameters_mapping`, plus every column of `units`, filled with NULLs when there is no `unit_id`. So for those rows, `uv.name` and `uv.short_name` are NULL, which makes the whole concatenated string NULL.

ana-berger:
I had a good chat with Bene and she was trying to help me solve it. The best way we found was to add a unitless unit to all the variables that have no unit

PR #841, opened 2020-07-30
SRS OC SOOP RAD - fix non harvesting of files

lbesnard:


lbesnard:
This PR is quite magical, really. 
I have no idea why the current version of the harvester on prod is not harvesting some files, and why is this one working. However, in this PR, by flattening subjobs into the main job, changing the old connection version of posgresql ... as well as a few other wizardry, the harvester is now working.
I'm always amazed at how Talend works. Even in 50 years, I still won't get it. 
This PR can be merged as blindly as when I modified the harvester

PR #843, opened 2020-07-30
PARAMETERS_MAPPING-new plank_prod view

ana-berger:


ana-berger:
example of the headers now:
[nrs_ind_map.txt](https://github.com/aodn/harvesters/files/4999984/nrs_ind_map.txt)

mhidas:
Looks better, but I don't like the "Unitless" units (plus there's a typo in the short version of this "unit"). I would prefer them to be blank. Did you try to use the `COALESCE` function in the query, which would allow the units columns to be null?

A couple more comments in the attached spreadsheet. 
[nrs_ind_map2.zip](https://github.com/aodn/harvesters/files/5013286/nrs_ind_map2.zip)

ana-berger:
I tried with the COALESCE, but it still makes the whole line empty, for the variables without unit_id. I agree it would look better empty than with unitless

bpasquer:
> I tried with the COALESCE, but it still makes the whole line empty, for the variables without unit_id. I agree it would look better empty than with unitless

Then you can use replace "unitless"  with  an empty string in the query

ana-berger:
> Looks better, but I don't like the "Unitless" units (plus there's a typo in the short version of this "unit"). I would prefer them to be blank. Did you try to use the `COALESCE` function in the query, which would allow the units columns to be null?

It is not a typo, I tried to make a short name for it, as for the other units. I followed the example of the 'Dimensionless' - 'Dmnless'. But maybe is does not look very good. I am still hoping we will be able to fix it and leave those cells empty. 

> A couple more comments in the attached spreadsheet.
> [nrs_ind_map2.zip](https://github.com/aodn/harvesters/files/5013286/nrs_ind_map2.zip)

I don't see any comments in this file, were they supposed to be highlighted or something?

ana-berger:
> > I tried with the COALESCE, but it still makes the whole line empty, for the variables without unit_id. I agree it would look better empty than with unitless
> 
> Then you can use replace "unitless" with an empty string in the query

Alright, trying to do that now.

ana-berger:
The COALESCE function actually worked. The problem was with the 'definition' column which was all empty. So removing that column fixed it. I will send the new view soon!

ana-berger:
Here is how it looks now:
[cpr_phyto_genus_map.txt](https://github.com/aodn/harvesters/files/5013941/cpr_phyto_genus_map.txt)

ana-berger:
@mhidas Can I also merge this?

mhidas:
@ana-berger The general idea is that if you put up a PR proposing some changes, then someone else reviews it, and merges it if they're happy.
Having said that, we do occasionally make an exception and merge our own PR, particularly in harvesters because they are a pain to review, or if it's a very small, trivial change. The Devs will probably tell you this is bad practice.

PR #844, opened 2020-08-03
clone of ABOS harvesters, new DWM

ana-berger:


ana-berger:
> Looks pretty good to me, but I haven't opened any of these in Talend yet. Have you tried that?

I tested the harvesters DWM_[CURRENTS, SOFS_SF, SOFS_SP, TS], they all generated the new views. Some views are empty because I am probably not using the correct nc files. I just guessed who they could be from the Portal and chose a random file for each harvester to test, but will have to do with the correct ones.

ana-berger:
All harvesters working with the changed data name files, I think we are good to merge it

ana-berger:
yay, very good, one thing worked with no conflicts or strange merges!

mhidas:
@lwgordonimos Could you please add these new harvesters on Jenkins so that they can be built?

- DWM_SOFS_SURFACE_FLUXES
- DWM_SOFS_SURFACE_PROPERTIES
- DWM_TS_SINGLE_INST_TIMESERIES

(Should take you 15 seconds, according to https://github.com/aodn/issues/issues/756#issuecomment-637210911)

lwgordonimos:
@mhidas those builds are now available in http://build.aodn.org.au.

I can confirm it took somewhere in the range of 15 seconds ± 2.47 seconds, plus overheads for lunch break, time to merge and deployment time.

PR #851, opened 2020-09-23
SOOP AUSCPR raw flat views

mhidas:
See https://github.com/aodn/PO-Backlog/issues/1123


mhidas:
As a side-effect, this fixes aodn/PO-Backlog#1

PR #852, opened 2020-10-06
update-spatial-extent-for-dwm_current-and-remove-abos

ana-berger:


mhidas (approved):
:+1: 
This is good as it is, but **NOT TO BE MERGED** until we have published all the renamed data!

PR #855, opened 2020-10-29
make dwm_currents view to include only dm files

ana-berger:


workspace/ANMN_METADATA/code/liquibase/anmn_metadata.changelog.xml
```diff
@@ -282,11 +282,11 @@
 
 	CREATE VIEW dwm_currents_map AS
           SELECT * FROM moorings_currents_map
-          WHERE url LIKE 'IMOS/DWM/%' ;
+          WHERE url LIKE 'IMOS/DWM/%' AND file_version='1' AND not realtime ;
```
mhidas, line 285: Actually, since both of these (`dwm_` and `anmn_`) are just subsets of `moorings_currents_map`, it would make more sense to add this clause to _that_ view, and then we don't need it again here.

mhidas (changes requested):
The change you made is correct, but
* I thought of a better way to do it; and
* You then need to copy this changelog back into the relevant Talend component, save the subjob, and commit those changes too. (Otherwise it will have no effect on the harvester)

mhidas (approved):
:+1:

PR #859, opened 2020-11-02
Fix liquibase script in ANMN_METADATA harvester

ana-berger:
please ignore the previous PR, other unrelated commits were included in that one


mhidas (approved):
My only comment is that I've renamed this PR to something a little more informative.

PR #862, opened 2020-11-30
remove ABOS harvesters directories

ana-berger:


mhidas:
:fireworks:

mhidas:
@lwgordonimos These harvesters can now be removed from build.aodn.org.au as well. Do we need a backlog item for that, or is it a 5-min job?

PR #864, opened 2020-12-14
new NRMN harvester

ana-berger:


bpasquer:
A couple of comments so far:

- in the Job iPostgresDBUpdate_1:the column `geom` is missing from the declaration for table Species_list

-  geom column are missing in all tPostgresqlOutput component, meaning that geom  values  are not harvested into target _map tables

ana-berger:
> A couple of comments so far:
> 
> * in the Job iPostgresDBUpdate_1:the column `geom` is missing from the declaration for table Species_list

so, this table does not have lat & lon data, should it have geom anyway?

> * geom column are missing in all tPostgresqlOutput component, meaning that geom  values  are not harvested into target _map tables

I am not sure it is the same thing or related, but we talked about adding the geom column in geoserver, from what I remember. We will talk about it.

Thanks Bene!

bpasquer:
> so, this table does not have lat & lon data, should it have geom anyway?

Yes, you're right, ignore this comment

ana-berger:
I made the changes; hope I did not forget anything

PR #872, opened 2021-01-12
working version of the harvester+pipeline in my stack

ana-berger:
one more attempt to run my working stack version in RC so I can add the most recent log when lodging an issue


mhidas (commented):
The only significant change I see is that you've added a 1-sec timeout to the database connection details (i'm not sure what the default is, otherwise). Is that what you needed to make it work in the PO stack?

PR #875, opened 2021-01-18
Animal tracking acoustic reporting new harvester

leonardolaiolo:
Loz, here is the new harvester created for https://github.com/aodn/PO-Backlog/issues/1506
To run in PO stack requires a new schema to be created, named ```animal_tracking_acoustic_installation```
It works fine for my understanding. 
In Talend I did not updated or deleted the last box 'UpdateMetadata' that I think it is not relevant at this stage but may be later on?


lbesnard:
you also need to add the externalization context,

see https://docs.google.com/document/d/1-E05wDMvI_Yy2Zz9m3D8DCWUjjkhPDJFm_jr9TvbUgc/edit

Most of the other harvest jobs use ‘Destination’ for the harvest database and ‘Externalisation’ for logDir and paramFile.

lbesnard:
@leonardolaiolo I rebased all the commits and removed your credentials so they don't appear in the repository publicaly

PR #884, opened 2021-02-03
SRS SST - decommission old harvester

lbesnard:
harvester should have been decommissioned a long time ago as not referenced any more in chef-private


mhidas:
Yay! Another bites the dust!

PR #958, opened 2022-05-23
delete realtime harvesters

ana-berger:


bpasquer (commented):
You went one step too far with the file deletion, we need to keep the current AODN_WAVE_NRT harvester

ana-berger:
> You went one step too far with the file deletion, we need to keep the current AODN_WAVE_NRT harvester

oops, I thought we didn't have a harvester for RT, so just removed them all. I will fix it

bpasquer:
> oops, I thought we didn't have a harvester for RT, so just removed them all. I will fix it

we don't have a pipeline since we are not ingesting files, but we do have a harvester that harvest data from the BOM WFS ;-)

ana-berger:
yes, I mixed things up. Actually I think I did a rm `-r AODN_WAVE_NRT*` and then all was gone

PR #976, opened 2022-10-11
(Fix) SRS SAR WIND - truncate measurements table and drop rows on every run

lbesnard:


lbesnard:
the measurements table was used as a temporary table to populate another table. The data was left in, but keeping it was not essential.
This PR truncates the table on the first run, then deletes every row from the measurements table at each run, using it essentially as a temporary table

PR #978, opened 2022-10-26
(Fix) SRS OC LJCO - fix poor performing map views

lbesnard:


lbesnard:
requires https://github.com/aodn/chef-private/pull/4252

bpasquer (approved):

PR #980, opened 2022-10-27
(Fix) SRS OC LJCO - add missing tables in data views used for filtering on step2

lbesnard:


lbesnard:
linked to:
- [ ] https://github.com/aodn/geoserver-config/pull/757

PR #987, opened 2022-11-28
feat: anmn_metadata new dwm_csiro_*_map

evacougnon:


mhidas (commented):
From what I can tell looking at just the XML (I haven't opened Talend in years...) it looks good, but I think there might be something missing - there was a small bit of SQL in the data harvest and delete subjobs that should be modified to set `DwmCsiroGriddedHourlyUpdated` to `true`, otherwise the harvester won't know when to update the spatial extent...

mhidas:
You had it in your previous PR:
https://github.com/aodn/harvesters/blob/060e5a499e4689a9560d925b85f4848197b8fb09/workspace/ANMN_METADATA/process/subjob/HarvestFileSystemMetadata_0.1.item#L889

evacougnon:
Good pick @mhidas thanks, it should have been part of the PR, I'm looking into it now

mhidas (approved):
:+1: Looks good now, as far as I can tell without opening the harvester in Talend :grin:

PR #990, opened 2022-12-07
Revert "disable aodn_wave_nrt spatial extent update"

bpasquer:
Reverts aodn/harvesters#989


bpasquer:
files from the AODN_WAVE harvester have made their way in the PR

PR #1008, opened 2023-04-06
Dwm csiro daily

evacougnon:
adding a new view to deal with daily and hourly csiro gridded DWM product in the same view/collection

Related geoserver PR: https://github.com/aodn/geoserver-config/pull/781


workspace/ANMN_METADATA/code/liquibase/anmn_metadata.changelog.xml
```diff
@@ -471,6 +471,36 @@
...
+                m.long_names,
+                m.site_code,
+                m.time_coverage_start AT TIME ZONE 'UTC' AS time_coverage_start,
+                m.time_coverage_end AT TIME ZONE 'UTC' AS time_coverage_end,
+                m.date_created AT TIME ZONE 'UTC' AS date_created,
+                i.first_indexed AT TIME ZONE 'UTC' AS date_published,
+                i.last_indexed AT TIME ZONE 'UTC' AS date_updated,
+                m.latitude,
+                m.longitude,
+                CASE WHEN i.url LIKE 'IMOS/DWM/DA/CSIRO_gridded_all_variables/%hourly-depth-gridded%' THEN 'hourly' ELSE 'daily' END AS data_type,
```
mhidas, line 493: Sorry, could we call this `product_type`? I think that's a better description (actually it doesn't matter too much what we call in in the db, but on the Portal I think "Product type" would be a better name for the filter).

Also, this assumes we'll only have 'hourly' or 'daily' and they won't come up with a third type later. I suggest something like this would be more flexible and future-proof...
```suggestion
                substring(url, 'FV02_([^_]*)_END') AS product_type,
```
So this will have values like "hourly-depth-gridded-product"
(For some reason I do also have an irrational aversion to CASE statements in SQL...)

mhidas (commented):
:+1: Looks good. Just one suggestion...

mhidas (approved):

